PR #1, opened 2021-10-11
add a basic prefix free compressor

kedartatwawadi:
Creating a sample Pull request to set the workflow amongst us. 
Typically we should check
- correctness
- if the code is well commented
- if there are sufficient tests
As this is mainly to help with research, we should really check if the comments/documentation is upto the mark.


shubhamchandak94:
A few thoughts on this:
- code organization - do we plan to put all the compressors in a separate directory? Should the FixedBitwidthCompressor compressor be moved there? 
- code organization - why is the BitsParserTransformer put in separate file than the other transforms?
- tests - we can have two types of tests - (i) roundtrip tests for small examples/edge cases and large random inputs, (ii) small tests where we check the compressed output wrt the expected compressed output derived by hand.

kedartatwawadi:
@shubhamchandak94:
1. We can do that, lets move `FixedBitWidthCompressor` to the compressor dir
2. I will add that transformer to the same file. Can think of better ways of handling the organization later
3. Regarding tests, that makes sense. I will create an issue for this, and we can slowly work on fixing the tests. 
- Random unit tests
- Fixed input unit tests

compressors/baseline_compressors.py
```diff
@@ -0,0 +1,73 @@
+"""
+Contains some elementary baseline compressors
```
kedartatwawadi, line 2: @shubhamchandak94: I moved all compressors to a single folder

kedartatwawadi:
Thanks @shubhamchandak94  for the comments! I have made the organizational changes. Let me know if everything else looks good (so that I can merge the PR)

shubhamchandak94 (approved):

compressors/baseline_compressors.py
```diff
@@ -0,0 +1,73 @@
...
+        # Example:
+        # data = [A,B,B,A,C]
+        # Transform-1: Lookuptable to get he alphabet index {A: 0, B: 1, C: 2}
+        # Transform-2: Uint to bitstring with a fixed width (2 for example) [0 -> 00, 1 -> 01, 2 -> 10]
+        # Transform-3: Bitstring to bits: [000 -> 0,0,0]
+        self.encoder_transform = CascadeTransformer(
+            [
+                LookupTableTransformer(self.encoder_lookup_table),
+                UintToBitstringTransformer(bit_width=self.bit_width),
+                BitstringToBitsTransformer(),
```
tpulkit, line 57: I am unclear on why you need BitstringToBitsTransformer? Is it similar to converting the block based codes to streaming codes? In that case, should we think of a general transform class going to-and-fro from "block-based" codes to "streamable" codes, e.g. as @shubhamchandak94 suggested for DataStream?

tpulkit (approved):
I just had one minor comment on the bitsToBitStream and vice-versa. I left it on the changed file. Let me know if there is a better way to communicate such comments. Thanks.

compressors/baseline_compressors.py
kedartatwawadi: @tpulkit, this is the recommended way of leaving comments. Typically we should leave the PR unmerged if there are any un-resolved comments!

kedartatwawadi: The final output of any encoder has to be a `BitsDataStream`, to be a valid encoder and to help with figuring out codelength etc. 

See the following snippet in `data_compressors.py`
```

    def encode(self, data_stream: DataStream):
        """
        The core encode function of the compressor
        """

        # set the parameters of the encoder/decoder using the data_stream
        self.set_encoder_decoder_params(data_stream)

        # perform the encoding
        output_bits_stream = self.encoder_transform.transform(data_stream)

        # the final output of encoder needs to be a stream of bits
        assert isinstance(output_bits_stream, BitsDataStream)
```

kedartatwawadi: This case, typically we will build a lookup take which will output a bitstring, so the transform just converts the bitstring to bits

PR #2, opened 2021-10-17
Update .gitignore

shubhamchandak94:
Ignore Python compiled files based on https://github.com/github/gitignore/blob/master/Python.gitignore


kedartatwawadi (approved):
Looks good!

Issue #6, opened 2021-10-20
Add random + non-random tests


kedartatwawadi:
-  [ ] Add random and fixed input tests for all compressors
-  [x] refactor the code to remove redundancy)

PR #7, opened 2021-10-22
Add implementation for huffman coding

kedartatwawadi:
1. Basing on `kedar/add_support_for_prefix_coding` as @tpulkit  is still reviewing the other PR. Will diff with master when that branch gets merged. 
2. Implemented a basic Huffman coding implementation
3. Cleanup of testing framework as suggested by @shubhamchandak94  (see #6). Still more cleanup + improvements to be done, but that we will do slowly


compressors/huffman_coder.py
```diff
@@ -0,0 +1,133 @@
...
+from core.data_transformer import (
+    BitstringToBitsTransformer,
+    CascadeTransformer,
+    BitsParserTransformer,
+    LookupTableTransformer,
+)
+
+
+@dataclass
+class HuffmanTree:
```
shubhamchandak94, line 15: should we call this class HuffmanNode, and have another class called HuffmanTree with a single variable called RootNode? That would be semantically more correct, but perhaps adds a bit more complexity

compressors/huffman_coder.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @property
+    def is_leaf_node(self):
+        return (self.left_child is None) and (self.right_child is None)
+
+    @classmethod
+    def build_huffman_tree(cls, prob_dist: ProbabilityDist):
+        """
+        Build the huffman coding tree
+        NOTE: Not the most efficient implementation. The insertion/sorting efficiency can be improved
```
shubhamchandak94, line 30: Should we switch to https://docs.python.org/3/library/heapq.html? It's clearly more efficient and might be conceptually equally as clear as the current implementation. One slight complication is in providing the sorting key to the heap thing. The best way is to have a custom comparator for the Huffman Node type (using an operator: https://docs.python.org/3/library/operator.html).
Overall, might be fine to leave it as is for now, but should be relatively straightforward to switch to a heap based implementation at some point.

compressors/huffman_coder.py
kedartatwawadi: Makes sense. Let me try to fix this

compressors/huffman_coder.py
```diff
@@ -0,0 +1,145 @@
...
+from core.data_transformer import (
+    BitstringToBitsTransformer,
+    CascadeTransformer,
+    BitsParserTransformer,
+    LookupTableTransformer,
+)
+
+
+@dataclass
+class HuffmanNode:
```
kedartatwawadi, line 15: @shubhamchandak94: Let me know if this makes more sense.

shubhamchandak94: Thanks, looks good

compressors/huffman_coder.py
kedartatwawadi: @shubhamchandak94 , I added this code!

compressors/huffman_coder.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def __le__(self, other):
+        """
+        Define a comparison operator, so that we can use this while comparing nodes
+        # NOTE: we only need to define one compare op, as others can be implemented using the
+        decorator @total_ordering
+        """
+        return self.prob <= other.prob
+
+    def _get_lines(self):
```
kedartatwawadi, line 61: @shubhamchandak94: this is the code I added ... please review!
I guess we should move to a generic Prefix code util instead of just huffman

shubhamchandak94:
Thanks, looks great! I added a super small commit with some extra print statements

shubhamchandak94 (approved):

PR #19, opened 2022-01-17
change "stream" to "block" in code since it more semantically correct

shubhamchandak94:


kedartatwawadi (approved):
Thanks for the change! Seems more logical now

tpulkit:
Looks better!

PR #20, opened 2022-01-17
add implementation of Golomb code

shubhamchandak94:


compressors/integer_coders.py
```diff
@@ -2,6 +2,7 @@
 from core.data_encoder_decoder import DataEncoder, DataDecoder
```
aysh-and-kedi, line 2: Wondering if it is better to just have one each compressor in a separate file. 
Probably that ll be better

compressors/integer_coders.py
```diff
@@ -85,3 +86,149 @@ def test_universal_uint_encode_decode():
...
+    if coder.cutoff == coder.M:
+        coder.rice_code = True
+    else:
+        coder.rice_code = False
+
+
+class GolombUintEncoder(DataEncoder):
+    """
+    Golomb code with parameter M based on
+    https://en.wikipedia.org/wiki/Golomb_coding#Simple_algorithm:
```
kedartatwawadi, line 110: Probably add a bit more of detail on the Golomb code algorithm here.

kedartatwawadi: Please use the google docstring style (probably add an add on the VSCode/atom) .. also please follow the same for the functions

compressors/integer_coders.py
```diff
@@ -85,3 +86,149 @@ def test_universal_uint_encode_decode():
...
     is_lossless, _ = try_lossless_compression(data_block, encoder, decoder)
     assert is_lossless
+
+
+"""
+shared initialization for Golomb encoder and decoder
+"""
+
+
+def _initialize_golomb_coder(coder, M: int):
```
kedartatwawadi, line 96: I think this is a good time to think about how we want to handle common params for encoders, decoders. 
1. One option is the way you have.. have a function which takes in a `coder` and initializes the params
2. Second option is to explicitly have an object called:
```
class GolombCodeParams:
    def __init__(self, M: ):
        self.M = M 
        self.b = ... 
        self.cutoff = ...
        self.is_rice_code = ...
 
 
 class GolombUintEncoder(DataEncoder):
     def __init__(self, M):
         self.params = GolombParams(M)
```

3. The third option is to have GolombUintEncoder subclass both (DataEncoder and GolombParams)
     Not very familiar with multiple inheritance, but probably should work

compressors/integer_coders.py
```diff
@@ -85,3 +86,149 @@ def test_universal_uint_encode_decode():
...
+    assert M > 0
+    coder.M = M
+    coder.b = int(math.floor(math.log2(coder.M)))
```
kedartatwawadi, line 99: I believe the names `M` and `b` are from wikipedia article.. but perhaps better to use more descriptive names or add a comment to what they mean

compressors/integer_coders.py
```diff
@@ -85,3 +86,149 @@ def test_universal_uint_encode_decode():
...
+    Golomb code with parameter M based on
+    https://en.wikipedia.org/wiki/Golomb_coding#Simple_algorithm:
+    If M is power of 2, we simplify to Rice codes with slight
+    change in logic.
+    """
+
+    def __init__(self, M: int):
+        _initialize_golomb_coder(self, M)
+
+    def encode_symbol(self, x: int):
```
kedartatwawadi, line 118: The VSCode add-on, should automatically add a very helpful docstring with the inputs/return etc. Do use that

compressors/integer_coders.py
```diff
@@ -85,3 +86,149 @@ def test_universal_uint_encode_decode():
...
+"""
+
+
+def _initialize_golomb_coder(coder, M: int):
+    assert M > 0
+    coder.M = M
+    coder.b = int(math.floor(math.log2(coder.M)))
+    coder.cutoff = 2 ** (coder.b + 1) - coder.M
+    if coder.cutoff == coder.M:
+        coder.rice_code = True
```
kedartatwawadi, line 102: can be simplified as `is_rice_code = (cutoff == M)`

compressors/integer_coders.py
```diff
@@ -85,3 +86,149 @@ def test_universal_uint_encode_decode():
...
+        num_bits_consumed = 0
+        while num_bits_consumed < len(bitarray):
+            s, num_bits = self.decode_symbol(bitarray[num_bits_consumed:])
+            num_bits_consumed += num_bits
+            data_list.append(s)
+
+        return DataBlock(data_list), num_bits_consumed
+
+
+def _test_golomb_encode_decode_helper(M, data_list, expected_output_bitarray):
```
kedartatwawadi, line 203: could use the "try_lossless..." utility function

compressors/integer_coders.py
shubhamchandak94: thanks, will do. If we could add the formatting expectations somewhere (Readme or in PR guidelines), that would be good. I just ran the black thing for now

compressors/integer_coders.py
shubhamchandak94: Not sure they have a name. I will add a comment.

compressors/integer_coders.py
shubhamchandak94: Do options 2 and 3 have any advantage over option 1? We can discuss more when we talk.

compressors/integer_coders.py
shubhamchandak94: Thanks, will do. Also see my comment below in another response:

> If we could add the formatting expectations somewhere (Readme or in PR guidelines), that would be good. I just ran the black thing for now

compressors/integer_coders.py
shubhamchandak94: Thanks, will separate it out into another file

compressors/integer_coders.py
shubhamchandak94: Just want to make sure we don't add complications without any reason.

compressors/integer_coders.py
kedartatwawadi: Yes, I think it might be in general useful.. can you change ` try_lossless_compression` to return the `encoded_bitarray` (instead of the length). and then we can check outside the function 

```
    is_lossless, encoded_bitarray = try_lossless_compression(data_block, encoder, decoder)
    assert is_lossless
    assert encoded_bitarray == expected_output_bitarray
```

This way, in case we don't want to test the `encoded_bitarray`, that output can be ignored.

compressors/integer_coders.py
kedartatwawadi: Yes ll do that

compressors/integer_coders.py
kedartatwawadi: sounds good

compressors/integer_coders.py
kedartatwawadi: yes, I agree we should definitely think if either of those options adds complications. 
(as this scenario of encoder/decoder sharing params is not uncommon) 

I like the option 2, as it makes things a bit more explicit. We could also use `@dataclass` to simplify a few things. 
Option 1, we are modifying an object, which felt less elegant to me.  Lets continue discussing

compressors/integer_coders.py
kedartatwawadi: https://gist.github.com/redlotus/3bc387c2591e3e908c9b63b97b11d24e FYI

compressors/integer_coders.py
kedartatwawadi: I use this add-on (suggested by @tpulkit) which makes it quite easy

```
Name: autoDocstring - Python Docstring Generator
Id: njpwerner.autodocstring
Description: Generates python docstrings automatically
Version: 0.6.1
Publisher: Nils Werner
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring
```

shubhamchandak94:
Made suggested changes. Also found and fixed a major mistake in the decoder implementation.

compressors/golomb_coder.py
```diff
@@ -0,0 +1,211 @@
+"""Golomb prefix coding for non-negative integers 
+
+Golomb code with parameter M based on
+https://en.wikipedia.org/wiki/Golomb_coding#Simple_algorithm.
+If M is power of 2, we simplify to Rice codes with slight
+change in logic.
+
+The main idea of Golomb code is to divide the input into quotient by M
```
kedartatwawadi, line 8: One small suggestion is to write this in points. Good that you included the decoding algo, as I didn't see that on the wikipedia link!

kedartatwawadi (approved):
Looks good! thanks for all the improvements :)

PR #23, opened 2022-02-13
PR for the framework changes

kedartatwawadi:
- [x] Added the changes to the encoder/decoder framework
- [x] Ported over the existing code to the new format
- [x] Added tests for the end-to-end pipeline (reading from a file, writing encoded stream to a file and then decoding the encoded stream) for the `FixedBitWidthEncoder`

As the PR is quite large, might be better to look at the branch itself. 

I would suggest: 
- Look at the baseline_compressor.py, integer_codes.py first to get an idea of usage
- Then look at the core folder to see the framework itself


core/data_stream.py
```diff
@@ -0,0 +1,249 @@
...
+    def write_block(self, data_block: DataBlock):
+        """write the input block to the stream
+
+        Args:
+            data_block (DataBlock): block to be written to the stream
+        """
+        for s in data_block.data_list:
+            self.write_symbol(s)
+
+    def __enter__(self):
```
shubhamchandak94, line 81: Nice! Learnt something new here. Maybe a short comment with a link to why this is useful can be handy for those new to Python

core/data_stream.py
```diff
@@ -0,0 +1,249 @@
...
+
+        Args:
+            block_size (int): the size of the block of data to be returned.
+
+        Returns:
+            DataBlock:
+        """
+        # returns the next data block
+        data_list = []
+        for _ in range(block_size):
```
shubhamchandak94, line 47: I like that we have a general implementation but I would recommend doing a more efficient implementation for both lists and files (using the slicing operation and the read functions respectively). I think that's the way someone would do it naturally if not trying to do something generic.

shubhamchandak94: same for write_block

core/data_stream.py
```diff
@@ -0,0 +1,249 @@
...
+    - get_block(block_size) -> returns a DataBlock of the given block_size from the stream
+    - write_block(block) -> writes the block of data to the stream
+
+    The DataStream can act as a stream object for both writing and reading blocks
+    The two more useful sub-classes of the abstract class are FileDataStream and ListDataStream.
+    (see their description for more details)
+    """
+
+    @abc.abstractmethod
+    def reset(self):
```
shubhamchandak94, line 24: TODO: explain the meaning of reset in more detail. In particular, if we reset a writeable stream what happens. For lists, it seems we move back the pointer but still keep appending to the end. Whereas for file, after reset we will overwrite.

compressors/baseline_compressors.py
```diff
@@ -3,71 +3,195 @@
...
+
+    is_lossless, codelen = try_lossless_compression(data_block, encoder, decoder)
+    assert is_lossless
+
+    # check if the length of the encoding was correct
+    alphabet_bits = (1 + len(data_block.get_alphabet())) * 8
+    assert codelen == len(data_list) * 2 + alphabet_bits
+
+
+def test_fixed_bitwidth_file_write():
```
shubhamchandak94, line 163: Should we move this to test_utils (similar to try_lossless_compression) since this is useful even beyond the fixed bitwidth compressor? And this is also how we would typically use the compressor, so having it outside the bitwidth file can make sense. 
One question I wonder about is whether encode_block/decode_block passing roundtrip implies that encode/decode roundtrip always passes. I think that's true but would like to confirm with you once.

utils/bitarray_utils.py
```diff
@@ -0,0 +1,36 @@
...
+# TODO: we could add more functions later
+class BitArray(bitarray.bitarray):
+    pass
+
+
+def uint_to_bitarray(x: int, bit_width=None) -> BitArray:
+    return BitArray(uint_to_bitstring(x, bit_width=bit_width))
+
+
+def bitarray_to_uint(bit_array: BitArray) -> int:
```
shubhamchandak94, line 34: not urgent or particularly relevant for efficiency, but can directly use ba2int (https://github.com/ilanschnell/bitarray)

shubhamchandak94 (commented):
Thanks for the incredible work on this! Left some comments. Happy to make some of the changes myself if you think that's better

compressors/baseline_compressors.py
kedartatwawadi: Yes, I will move this to utils.
In general encode_block/decode_block roundtrip would imply encode/decode always passes. We however need to make sure the "DataStream" we are using is able to read/write correctly. So probably need a test for the DataStream as well. 

Also need to think how to abstract this correctly for different datastreams ... (we will most commonly have TextDataStream, but we could have others.. so need to think about that)

core/data_stream.py
```diff
@@ -0,0 +1,249 @@
...
+    @abc.abstractmethod
+    def get_symbol(self):
+        """returns a symbol from the data stream, returns None if the stream is finished
+
+        This is an abstract method, and hence needs to be implemented by the subclasses
+        """
+        pass
+
+    def get_block(self, block_size: int) -> DataBlock:
+        """returns a block of data (of the given max size) from the stream
```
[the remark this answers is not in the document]
kedartatwawadi: Done

core/data_stream.py
kedartatwawadi: Thanks! done

core/data_stream.py
kedartatwawadi: - The difference in implementation of reset for List vs File is a bug. 
- Now that I think of it "reset" doesn't make sense while writing. It probably makes sense while reading. For example: what does it mean to "reset" a file in "w" mode vs "a" mode. 

There are two ways of dealing with this: 
a. Instead of defining a `reset` function.. we define a `seek` function to go to a particular symbol location. So `seek(0)` takes pointer to the beginning of the list/File, while `seek(k)` takes it to the location of the `kth` symbol. 
(this implementation could vary based on whether we are reading integers/characters etc.). 
This will remove the ambiguity. It also seems consistent with the default python file object which doesn't have a 'reset' method, but ahs a `seek` method. 

b) The second method is defining a separate File Writer and Reader and defining `reset` only for the Reader. 

Let me know your thoughts on either of these. I think the first option is a decent one IMO

core/data_stream.py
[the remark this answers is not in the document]
shubhamchandak94: Thanks!

core/data_stream.py
shubhamchandak94: Thanks!

utils/bitarray_utils.py
shubhamchandak94: 👍

core/data_stream.py
shubhamchandak94: Makes sense. As long as we don't mark this as @final there is no hurry to do this. Can just leave a TODO comment somewhere.

core/data_stream.py
shubhamchandak94: First option sounds good. Will seek make sense when writing? If not, we should at least add a comment to clarify.

compressors/baseline_compressors.py
shubhamchandak94: Perhaps some way to pass the stream type to the test function?

core/data_stream.py
kedartatwawadi: Done!

kedartatwawadi: Seek probably makes sense for writing, if we want to modify a particular symbol

core/data_stream.py
kedartatwawadi: done! added a NOTE and a TODO for now

shubhamchandak94: Thanks. Looks good.

core/data_stream.py
shubhamchandak94: The code looks correct now for both list and file, thanks for adding the tests.

compressors/baseline_compressors.py
kedartatwawadi: I refactored this part of the code. @shubhamchandak94 @tpulkit 
- Added utility function `encode_file` and `decode_file` to DataEncoder and DataDecoder. This is a bit of a change to the API, so do take a look and let me know if this makes sense. I am not 100% sure about this, so would be good to take a look and see if there is a way to improve.
- Added utility functions to `test_utils.py`

core/data_encoder_decoder.py
```diff
@@ -0,0 +1,175 @@
...
+                break
+
+            # encode the block
+            output = self.encode_block(data_block)
+            assert isinstance(output, BitArray)
+
+            # write encoded bitarrays
+            encode_writer.write_block(output)
+
+    def encode_file(self, input_file_path: str, encoded_file_path: str, block_size: int = 10000):
```
kedartatwawadi, line 79: @shubhamchandak94 -> new addition since your comments

shubhamchandak94 (approved):

PR #31, opened 2022-03-10
Add Shannon Fano Elias code implementation

kedartatwawadi:


compressors/shannon_fano_elias_coder.py
```diff
@@ -0,0 +1,94 @@
...
+import math
+import numpy as np
+
+
+class ShannonFanoEliasEncoder(PrefixFreeEncoder):
+    def __init__(self, prob_dist: ProbabilityDist):
+        self.prob_dist = prob_dist
+
+        # compute a dictionary holding cumulative probabilities
+        self.cumulative_prob_dict = self.prob_dist.cumulative_prob_dict
```
shubhamchandak94, line 24: is this a fn call to compute the cumprob?

compressors/shannon_fano_elias_coder.py
```diff
@@ -0,0 +1,94 @@
...
+            # check if range_end and range_start are both in the same bucket
+            cum_prob_list = list(self.cumulative_prob_dict.values())
+            start_bin = np.searchsorted(cum_prob_list, range_start, side="right")
+            end_bin = np.searchsorted(cum_prob_list, range_end, side="left")
+
+            if start_bin == end_bin:
+                decoded_symbol = self.prob_dist.alphabet[start_bin - 1]
+
+                # FIXME: The recomputing of num_bits_consumed seems necessary
+                num_bits_consumed = math.ceil(self.prob_dist.log_probability(decoded_symbol)) + 1
```
shubhamchandak94, line 72: I think I understand what you mean here (basically the true num_bits_consumed can be bigger than the minimum number of bits needed to get completely within the range due to some suboptimality in the code). Might be good to elaborate here so we can understand later on

shubhamchandak94 (commented):
Left a couple comments in code. More broadly, do you think there is some scope for error because of finite precision float arithmetic (e.g., when adding to find cumsum or when converting to bitarray by multiplying by 2^precision or when taking logs)? I don't think this is a priority to fix (as opposed to range coder which we want to get perfect since it is used a lot), but can add a note if we think this might not work for things like super small probabilities

compressors/shannon_fano_elias_coder.py
kedartatwawadi: yes that is precissely what I want to say. This is in fact an issue in Arithmetic coding as well. (and is one thing which is not resolved by my code yet)

kedartatwawadi:
@shubhamchandak94: Broadly I do think there could be issues due to finite precision. I added `Frequencies` etc. to the arithmetic coding code, so can re-write Shannon-Fano using those to avoid floating point issues + terribly small probability issues. 

I will think and get back

compressors/shannon_fano_elias_coder.py
kedartatwawadi: yes it is a function call.. but if we decorate it with `@property`, it can be viewed as an attribute + it gets cached for efficiency

compressors/shannon_fano_elias_coder.py
kedartatwawadi: i will add a comment with that effect in the prod_dist code

compressors/shannon_fano_elias_coder.py
kedartatwawadi: i will add a comment with that effect in the prob_dist code

compressors/shannon_fano_elias_coder.py
kedartatwawadi: elaborated the comment

core/prob_dist.py
```diff
@@ -45,7 +61,7 @@ def _validate_prob_dist(prob_dict):
 
         sum_of_probs = 0
         for _, prob in prob_dict.items():
-            assert prob >= 0, "probabilities cannot be negative"
+            assert prob >= 1e-6, "probabilities negative or too small cause stability issues"
```
kedartatwawadi, line 64: @tpulkit

shubhamchandak94 (approved):

kedartatwawadi:
@tpulkit: merging this.. we can re-visit some parts if needed

tpulkit:
sounds good.

PR #35, opened 2022-04-04
Added Shannon-Fano Codes

tpulkit:
branches off Kedar's Shannon-Fano-Elias and uses some functions (particularly utils/prob_dist) to build the Shannon coder. Might make more sense to first merge that branch and then this.

### DIFF
--- a/compressors/fano_coder.py
+++ b/compressors/fano_coder.py
@@ -1,4 +1,4 @@
-"""Shannon Coding
+"""Fano Coding
 Shannon coding is a basic prefix free coding. There is some confusion in literature about Shannon, Fano and
 Shannon-Fano codes, e.g. see Naming section of https://en.wikipedia.org/wiki/Shannon–Fano_coding.
 
@@ -10,10 +10,13 @@ from utils.test_utils import get_random_data_block, try_lossless_compression
 from compressors.prefix_free_compressors import PrefixFreeTree, PrefixFreeEncoder, PrefixFreeDecoder, BinaryNode
 from core.prob_dist import ProbabilityDist
 
+
 class FanoTree(PrefixFreeTree):
     def __init__(self, prob_dist):
         super().__init__(prob_dist)
+        # sort the probability distribution
         self.sorted_prob_dist = ProbabilityDist.get_sorted_prob_dist(prob_dist.prob_dict)
+        # initialize root node of Fano Tree
         self.root_node = BinaryNode(id=None)
 
         # build tree returns the root node
@@ -22,40 +25,41 @@ class FanoTree(PrefixFreeTree):
 
     @staticmethod
     def normalize_prob_dict(prob_dict):
+        """
+        normalizes subset of original probability dict to 1
+        """
         sum_p = sum(prob_dict.values())
-        return ProbabilityDist(dict([[a, b/sum_p] for a, b in prob_dict.items()]))
+        return ProbabilityDist(dict([[a, b / sum_p] for a, b in prob_dict.items()]))
 
     @staticmethod
     def criterion(dict_item):
+        """used for finding the symbol at which cumulative probability is nearest to 0.5
+        This symbol is used to split between left and right trees of Fano. We split right starting this symbol
+        to get most balanced tree.
+        """
         key, value = dict_item
         return abs(value - 0.5)
 
     @staticmethod
     def build_fano_tree(root_node, norm_sort_prob_list) -> BinaryNode:
+        """recursively build Fano Tree"""
+        # Get cumulative probability dict
         cumulative_prob_dict = norm_sort_prob_list.cumulative_prob_dict
 
         left_prob_dict = {}
         right_prob_dict = {}
 
+        # IF only two symbols left, we don't need to recurse further. These iwll be left and right child of current node
         if norm_sort_prob_list.size == 2:
             left_symbol, right_symbol = norm_sort_prob_list.alphabet
             root_node.left_child = BinaryNode(id=left_symbol)
             root_node.right_child = BinaryNode(id=right_symbol)
             return root_node
 
-        # prev_diff = 0
-        # for s, cum_prob in cumulative_prob_dict.items():
-        #     if cum_prob < 0.5:
-        #         left_prob_dict.update({s: norm_sort_prob_list.probability(s)})
-        #     else:
-        #         curr_diff = abs(0.5 - cum_prob)
-        #         if prev_diff > curr_diff:
-        #             left_prob_dict.update({s: norm_sort_prob_list.probability(s)})
-        #         else:
-        #             right_prob_dict.update({s: norm_sort_prob_list.probability(s)})
-        #
-        #     prev_diff = abs(0.5 - cum_prob)
-
+        # ELSE split the tree into left and right nodes based on cumulative probability and recursively build
+        # FanoTree for these trees
+        # all alphabets which are nearest to 0.5 cumulative probability in sorted list are put to left tree and rest
+        # to right, such that we have a most balanced probabilistic tree subsets.
         min_diff_code = min(cumulative_prob_dict.items(), key=FanoTree.criterion)[0]
         flag = "left"
         for s, cum_prob in cumulative_prob_dict.items():
@@ -66,6 +70,8 @@ class FanoTree(PrefixFreeTree):
             else:
                 right_prob_dict.update({s: norm_sort_prob_list.probability(s)})
 
+        # Call recursion
+        # if only 1 symbol in either left or right tree, just assign it as a child and we don't have to call recursion
         if len(left_prob_dict) != 1:
             root_node.left_child = BinaryNode(id=None)
             FanoTree.build_fano_tree(root_node.left_child, FanoTree.normalize_prob_dict(left_prob_dict))
@@ -84,8 +90,11 @@ class FanoTree(PrefixFreeTree):
 
 class FanoEncoder(PrefixFreeEncoder):
     """
-    Outputs codewords
+    PrefixFreeEncoder already has a encode_block function to encode the symbols once we define a encode_symbol function
+    for the particular compressor.
+    PrefixFreeTree provides get_encoding_table given a PrefixFreeTree
     """
+
     def __init__(self, prob_dist: ProbabilityDist):
         tree = FanoTree(prob_dist)
         self.encoding_table = tree.get_encoding_table()
@@ -95,6 +104,12 @@ class FanoEncoder(PrefixFreeEncoder):
 
 
 class FanoDecoder(PrefixFreeDecoder):
+    """
+    PrefixFreeDecoder already has a decode_block function to decode the symbols once we define a decode_symbol function
+    for the particular compressor.
+    PrefixFreeTree provides decode_symbol given a PrefixFreeTree
+    """
+
     def __init__(self, prob_dist: ProbabilityDist):
         self.tree = FanoTree(prob_dist)
 
@@ -104,18 +119,26 @@ class FanoDecoder(PrefixFreeDecoder):
 
 
 def test_fano_coding():
-    """test if shannon fano encoding is lossless"""
     NUM_SAMPLES = 2000
     distributions = [
         ProbabilityDist({"A": 0.5, "B": 0.5}),
         ProbabilityDist({"A": 0.3, "B": 0.3, "C": 0.4}),
-        ProbabilityDist({"A": 0.3, "B": 0.3, "C": 0.3, "D": 0.1}),
         ProbabilityDist({"A": 0.5, "B": 0.25, "C": 0.12, "D": 0.13}),
         ProbabilityDist({"A": 0.9, "B": 0.1})
     ]
-    for prob_dist in distributions:
+    expected_codewords = [
+        {"A": BitArray('0'), "B": BitArray('1')},
+        {"A": BitArray('10'), "B": BitArray('11'), "C": BitArray('0')},
+        {"A": BitArray('0'), "B": BitArray('10'), "C": BitArray('111'), "D": BitArray('110')},
+        {"A": BitArray('0'), "B": BitArray('1')}
+    ]
+
+    def test_end_to_end(prob_dist, num_samples):
+        """
+        Test if decoding of (encoded symbol) results in original
+        """
         # generate random data
-        data_block = get_random_data_block(prob_dist, NUM_SAMPLES, seed=0)
+        data_block = get_random_data_block(prob_dist, num_samples, seed=0)
 
         # create encoder decoder
         encoder = FanoEncoder(prob_dist)
@@ -124,3 +147,16 @@ def test_fano_coding():
         # perform compression
         is_lossless, _, _ = try_lossless_compression(data_block, encoder, decoder)
         assert is_lossless, "Lossless compression failed"
+
+    def test_encoded_symbol(prob_dist, expected_codeword_dict):
+        """
+        test if the encoded symbol is as expected
+        """
+        encoder = FanoEncoder(prob_dist)
+        print(encoder.encoding_table)
+        for s in prob_dist.prob_dict.keys():
+            assert encoder.encode_symbol(s) == expected_codeword_dict[s]
+
+    for i, prob_dist in enumerate(distributions):
+        test_end_to_end(prob_dist, NUM_SAMPLES)
+        test_encoded_symbol(prob_dist, expected_codeword_dict=expected_codewords[i])
--- a/compressors/fano_coder.py
+++ b/compressors/fano_coder.py
@@ -1,0 +1,95 @@
+"""Shannon Coding
+Shannon coding is a basic prefix free coding. There is some confusion in literature about Shannon, Fano and
+Shannon-Fano codes, e.g. see Naming section of https://en.wikipedia.org/wiki/Shannon–Fano_coding.
+
+This document uses Fano coding as described in the wiki article above.
+"""
+from utils.bitarray_utils import float_to_bitarrays, BitArray, uint_to_bitarray
+from utils.test_utils import get_random_data_block, try_lossless_compression
+from compressors.prefix_free_compressors import PrefixFreeTree, PrefixFreeTreeNode, PrefixFreeTreeEncoder, PrefixFreeTreeDecoder
+from core.prob_dist import ProbabilityDist
+import math
+
+
+class FanoTree(PrefixFreeTree):
+    def __init__(self, prob_dist):
+        super().__init__(prob_dist)
+        self.sorted_prob_dist = prob_dist.sorted_prob_list
+        self.root_node = PrefixFreeTreeNode(id=None)
+
+        # build tree returns the root node
+        self.build_tree(self.root_node, self.sorted_prob_dist)
+
+    @staticmethod
+    def normalize_prob_dict(prob_dict):
+        sum_p = sum(prob_dict.values())
+        return ProbabilityDist(dict([[a, b/sum_p] for a, b in prob_dict.items()]))
+
+    @staticmethod
+    def build_tree(root_node, norm_sort_prob_list) -> PrefixFreeTreeNode:
+        cumulative_prob_dict = norm_sort_prob_list.cumulative_prob_dict
+
+        left_prob_dict = {}
+        right_prob_dict = {}
+
+        if norm_sort_prob_list.size == 2:
+            left_symbol, right_symbol = norm_sort_prob_list.alphabet
+            root_node.left_child = PrefixFreeTreeNode(id=left_symbol)
+            root_node.right_child = PrefixFreeTreeNode(id=right_symbol)
+            return root_node
+
+        for s, cum_prob in cumulative_prob_dict.items():
+            if cum_prob < 0.5:
+                left_prob_dict.update({s: norm_sort_prob_list.probability(s)})
+            else:
+                right_prob_dict.update({s: norm_sort_prob_list.probability(s)})
+
+        if len(left_prob_dict) != 1:
+            root_node.left_child = PrefixFreeTreeNode(id=None)
+            FanoTree.build_tree(root_node.left_child, FanoTree.normalize_prob_dict(left_prob_dict))
+        else:
+            root_node.left_child = PrefixFreeTreeNode(id=list(left_prob_dict)[0])
+
+        if len(right_prob_dict) != 1:
+            root_node.right_child = PrefixFreeTreeNode(id=None)
+            FanoTree.build_tree(root_node.right_child, FanoTree.normalize_prob_dict(right_prob_dict))
+        else:
+            root_node.right_child = PrefixFreeTreeNode(id=list(right_prob_dict)[0])
+
+        return root_node
+
+
+class FanoEncoder(PrefixFreeTreeEncoder):
+    """
+    Outputs codewords
+    """
+
+    def __init__(self, prob_dist: ProbabilityDist):
+        self.tree = FanoTree(prob_dist)
+
+
+class FanoDecoder(PrefixFreeTreeDecoder):
+    def __init__(self, prob_dist: ProbabilityDist):
+        self.tree = FanoTree(prob_dist)
+
+
+def test_fano_coding():
+    """test if shannon fano encoding is lossless"""
+    NUM_SAMPLES = 2000
+    distributions = [
+        ProbabilityDist({"A": 0.5, "B": 0.5}),
+        ProbabilityDist({"A": 0.3, "B": 0.3, "C": 0.4}),
+        ProbabilityDist({"A": 0.5, "B": 0.25, "C": 0.12, "D": 0.13}),
+        ProbabilityDist({"A": 0.9, "B": 0.1})
+    ]
+    for prob_dist in distributions:
+        # generate random data
+        data_block = get_random_data_block(prob_dist, NUM_SAMPLES, seed=0)
+
+        # create encoder decoder
+        encoder = FanoEncoder(prob_dist)
+        decoder = FanoDecoder(prob_dist)
+
+        # perform compression
+        is_lossless, _, _ = try_lossless_compression(data_block, encoder, decoder)
+        assert is_lossless, "Lossless compression failed"
--- a/compressors/fano_coder.py
+++ b/compressors/fano_coder.py
@@ -14,9 +14,9 @@ from core.prob_dist import ProbabilityDist
 
 class FanoTree(PrefixFreeTree):
     def __init__(self, prob_dist):
-        super().__init__(prob_dist)
+        self.prob_dist = prob_dist
         # sort the probability distribution
-        self.sorted_prob_dist = ProbabilityDist.get_sorted_prob_dist(prob_dist.prob_dict)
+        self.sorted_prob_dist = ProbabilityDist.get_sorted_prob_dist(prob_dist.prob_dict, descending=True)
         # initialize root node of Fano Tree
         self.root_node = BinaryNode(id=None)
 
--- a/compressors/fano_coder.py
+++ b/compressors/fano_coder.py
@@ -22,13 +22,6 @@ class FanoTree(PrefixFreeTree):
         # build tree returns the root node
         super().__init__(root_node=self.build_fano_tree(self.root_node, self.sorted_prob_dist))
 
-    @staticmethod
-    def normalize_prob_dict(prob_dict):
-        """
-        normalizes subset of original probability dict to 1
-        """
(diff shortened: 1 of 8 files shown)


core/prob_dist.py
```diff
@@ -29,6 +29,11 @@ def alphabet(self):
     def prob_list(self):
         return [self.prob_dict[s] for s in self.alphabet]
 
+    @property
+    def sorted_prob_list(self):
```
kedartatwawadi, line 33: Please add a docstring, and maybe a quick test to check this

compressors/prefix_free_compressors.py
```diff
@@ -122,7 +122,8 @@ def __init__(self, prob_dist: ProbabilityDist):
         """
         create the prefix free tree
         """
-        self.root_node = self.build_tree(prob_dist)
+        self.prob_dist = prob_dist
+        self.root_node = PrefixFreeTreeNode(id=None, code=None)
```
kedartatwawadi, line 126: Please add a comment something like: 
```
Initialize the root node to an empty node
NOTE: the tree will need to be constructed by the Subclassing code and the root_node will need to be set to the correct node. 
```

compressors/prefix_free_compressors.py
```diff
@@ -134,16 +135,62 @@ def print_tree(self):
     def build_tree(prob_dist) -> PrefixFreeTreeNode:
         """
         abstract function -> needs to be implemented by the subclassing class
+
+        needs to build the whole tree and return the pointer to root node
         """
         raise NotImplementedError
 
+    def build_tree_from_code(self, symbol, code) -> PrefixFreeTreeNode:
```
kedartatwawadi, line 143: Consider making this function a "private function" by renaming it with `_build...` ..
Also, maybe rename this: `_add_code_to_the_tree` or something similar which might be more appropriate

compressors/prefix_free_compressors.py
```diff
@@ -134,16 +135,62 @@ def print_tree(self):
     def build_tree(prob_dist) -> PrefixFreeTreeNode:
```
kedartatwawadi, line 135: maybe just remove this function from `PrefixFreeTree`.. HuffmanTree can still define the function.

utils/bitarray_utils.py
```diff
@@ -15,7 +15,7 @@ def get_bit_width(size) -> int:
 
 def uint_to_bitarray(x: int, bit_width=None) -> BitArray:
     """
-    converts an unsigned into to bits.
+    converts an unsigned int to bits.
```
kedartatwawadi, line 18: thanks!

compressors/prefix_free_compressors.py
kedartatwawadi: Also complete the docstring!

compressors/prefix_free_compressors.py
```diff
@@ -122,7 +122,8 @@ def __init__(self, prob_dist: ProbabilityDist):
         """
         create the prefix free tree
         """
-        self.root_node = self.build_tree(prob_dist)
+        self.prob_dist = prob_dist
+        self.root_node = PrefixFreeTreeNode(id=None, code=None)
 
     def get_encoding_table(self):
```
kedartatwawadi, line 128: I had added this code for encoding_table.. but as you mentioned yday, it is un-intuitive to have this a function of the `PrefixFreeNode`. I will check later if this can be fixed and moved here, as that is more that.

compressors/shannon_coder.py
```diff
@@ -0,0 +1,90 @@
...
+        for s in prob_dist.prob_dict.keys():
+            encoding_table.update({s: self.get_codeword(s)})
+
+        print(f"Decoder: Encoding Table = {encoding_table}")
+
+        self.tree.build_tree_from_encoding_table(encoding_table)
+
+
+def test_shannon_coding():
+    """test if shannon fano elais encoding is lossless"""
```
kedartatwawadi, line 72: fix the comment

compressors/shannon_coder.py
```diff
@@ -0,0 +1,90 @@
+"""Shannon Coding
+Shannon coding is a basic prefix free coding. There is some confusion in literature about Shannon, Fano and
```
kedartatwawadi, line 2: typo: .. is a basic prefix free code.

compressors/prefix_free_compressors.py
kedartatwawadi: https://github.com/kedartatwawadi/stanford_compression_library/issues/38 added an issue for this

core/prob_dist.py
```diff
@@ -32,6 +32,14 @@ def alphabet(self):
     def prob_list(self):
         return [self.prob_dict[s] for s in self.alphabet]
 
+    @classmethod
+    def get_sorted_prob_dist(cls, prob_dict, reverse=True):
+        """
+        Returns:
+            an inheritable ProbabilityDist object with sorted probabilities
```
kedartatwawadi, line 39: didn't understand the "inheritable ProbabilityDist" part of the comment.what do you mean?

compressors/shannon_coder.py
```diff
@@ -0,0 +1,178 @@
...
+        """ function to generate prefix-free tree from code.
+        Args:
+            symbol: current symbol
+            code: current code
+            root_node: root node to the ShannonTree
+
+        Returns:
+            the pointer to root node of the tree so far
+        """
+        code_so_far = BitArray()
```
kedartatwawadi, line 60: A couple of suggestions here: 
1. Try to reduce code repetition, as that makes code longer + can lead to bugs. 
2. Add inline comments while you code, so that you don't need to add them later + it might help your thought process. 

Also, here is a suggestion for changing you code.. you might have to get it working, as I have not tested this: 
```python
# initialize the curr_node, code_so_far temporary var
curr_node = self.root_node
code_so_far = BitArray()

for i, bit in enumerate(code):
    code_so_far.append(bit)

    # get a pointer to the child node 
    child = curr_node.right_child if bit else curr_node.left_child

    if child is None:
        # add a new node, in case it is not already present
        new_node = ...
        child = new_node
    curr_node = child
```

compressors/shannon_coder.py
```diff
@@ -0,0 +1,90 @@
+"""Shannon Coding
+Shannon coding is a basic prefix free coding. There is some confusion in literature about Shannon, Fano and
+Shannon-Fano codes, e.g. see Naming section of https://en.wikipedia.org/wiki/Shannon–Fano_coding.
+
```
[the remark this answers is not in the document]
kedartatwawadi: @tpulkit: when you get some time feel free to update this short wiki blurb

compressors/shannon_coder.py
kedartatwawadi: @tpulkit: re-pasting my comment, as that one became outdated

core/prob_dist.py
```diff
@@ -32,6 +32,13 @@ def alphabet(self):
     def prob_list(self):
         return [self.prob_dict[s] for s in self.alphabet]
 
+    @classmethod
+    def get_sorted_prob_dist(cls, prob_dict, reverse=True):
+        """
+        Returns ProbabilityDist class object with sorted probabilities
```
kedartatwawadi, line 38: it would be good to clarify the default sort order (i.e. increasing vs decreasing... the reverse flag is not very intuitive

compressors/shannon_coder.py
```diff
@@ -0,0 +1,155 @@
...
+
+class ShannonTree(PrefixFreeTree):
+    """
+    Generates Codewords based on Shannon Coding. Allows generating PrefixFreeTree from the obtained codewords.
+    """
+
+    def __init__(self, prob_dist: ProbabilityDist):
+        self.prob_dist = prob_dist
+        # sort the probability distribution and get cumulative probability which will be used for encoding
```
kedartatwawadi, line 22: can you please edit as.. `#sort in the decreasing order the probability...`

compressors/fano_coder.py
```diff
@@ -0,0 +1,154 @@
...
+    def criterion(dict_item):
+        """used for finding the symbol at which cumulative probability is nearest to 0.5
+        This symbol is used to split between left and right trees of Fano. We split right starting this symbol
+        to get most balanced tree.
+        """
+        key, value = dict_item
+        return abs(value - 0.5)
+
+    @staticmethod
+    def build_fano_tree(root_node, norm_sort_prob_list) -> BinaryNode:
```
kedartatwawadi, line 43: Please fix variable name to be `norm_sort_prob_dist`

compressors/fano_coder.py
```diff
@@ -0,0 +1,154 @@
...
+    @staticmethod
+    def normalize_prob_dict(prob_dict):
+        """
+        normalizes subset of original probability dict to 1
+        """
+        sum_p = sum(prob_dict.values())
+        return ProbabilityDist(dict([[a, b / sum_p] for a, b in prob_dict.items()]))
+
+    @staticmethod
+    def criterion(dict_item):
```
kedartatwawadi, line 34: Instead of the function "criterion" which doesnt make much logical sense, 
probably move some of the prob_dist logic and create the following func:
```
    def split_prod_dist_into_two(prob_dist):
        # find the halfway split index
        split_index = min(prob_dist.cumulative_prob_dict.values(), key=lambda x: abs(x - 0.5))

        # split prob_dist into two based on split_index
        ...

        # normalize both of the dicts
        return left_prob_dist, right_prob_dist
```

compressors/fano_coder.py
```diff
@@ -0,0 +1,154 @@
...
+        key, value = dict_item
+        return abs(value - 0.5)
+
+    @staticmethod
+    def build_fano_tree(root_node, norm_sort_prob_list) -> BinaryNode:
+        """recursively build Fano Tree"""
+        # Get cumulative probability dict
+        cumulative_prob_dict = norm_sort_prob_list.cumulative_prob_dict
+
+        # Split the tree into left and right nodes based on cumulative probability and recursively build
```
kedartatwawadi, line 48: Once you move some logic to `split_prob_dist_into_two`, this function can be simplified as:

```python
    def build_fano_tree(root_node, norm_sort_prob_dist):
        # base case: 
        if only one node: 
           root_node.id = ...
           return root_node
        else:
           # split norm_sort_prob_dist into two normalized prob dists
           left_prob_dist, right_prob_dist = split_prod_dist_into_two(norm_sort_prob_dist)
           
            # call recursion on left node
           root_node.left_node = BinaryNode(id="None", ...)
           build_fano_tree(root_node.left_node, left_prob_dist)
           
           # call recursion on right node
           root_node.left_node = BinaryNode(id="None", ...)
           build_fano_tree(root_node.left_node, left_prob_dist)
           
           return root_node
        
```

kedartatwawadi: I have not checked whether this works, but you can try to get it working..
In general, it is better to start with writing a pseudo-code like structure and then filling in the details

compressors/fano_coder.py
```diff
@@ -0,0 +1,154 @@
...
+        # sort the probability distribution
+        self.sorted_prob_dist = ProbabilityDist.get_sorted_prob_dist(prob_dist.prob_dict)
+        # initialize root node of Fano Tree
+        self.root_node = BinaryNode(id=None)
+
+        # build tree returns the root node
+        super().__init__(root_node=self.build_fano_tree(self.root_node, self.sorted_prob_dist))
+
+    @staticmethod
+    def normalize_prob_dict(prob_dict):
```
kedartatwawadi, line 26: this is a useful general function.. maybe move it to prob_dist: 

```
def normalize_dict_and_create_prob_dist(dict):
     """
     normalizes dict -> dict_norm so that the sum of values is 1
     wraps dict_norm as a ProbabilityDist
     """
```

kedartatwawadi (approved):
Looks good overall! Thanks for all the hard work :) 
I had some suggestions for cosmetic changes to the `fano_coder`

compressors/shannon_coder.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # compute encode length
+        encode_len = math.ceil(self.prob_dist.log_probability(symbol))
+
+        # the encode is the binary representation of the cumulative probability in ascending order,
+        _, code = float_to_bitarrays(cum_prob, encode_len)
+        return code
+
+    @staticmethod
+    def build_tree_from_code(symbol, code, root_node) -> BinaryNode:
```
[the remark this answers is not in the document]
tpulkit: Not updating because of following reasons:

- The function is building a tree starting from a code and not adding just code to the tree. I have made a private function though by adding an underscore
- We need to return root as we recursively build tree starting from each node. See build_shannon_tree function

compressors/shannon_coder.py
```diff
@@ -0,0 +1,155 @@
...
+            code: current code
+            root_node: root node to the ShannonTree
+
+        Returns:
+            the pointer to root node of the tree so far
+        """
+        # initialize the curr_node, code_so_far temporary var
+        curr_node = root_node
+        code_so_far = BitArray()
+        code_len = len(code)
```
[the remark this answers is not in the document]
tpulkit: Thanks.
I will leave it for two reasons:
- just easier on eye
- why do we want to calculate len() again and again in the condition for the for loop where this variable is used -- very minor but still.

compressors/fano_coder.py
tpulkit: this has some issue. happy to discuss during our call today.

compressors/shannon_coder.py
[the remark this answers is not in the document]
kedartatwawadi: Cool! Yeah, it is fine either way. btw, `len(.)` doesn't really calculate the length. It just returns an internal variable.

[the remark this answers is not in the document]
tpulkit: good to know!

compressors/fano_coder.py
```diff
@@ -0,0 +1,160 @@
+"""Fano Coding
+Shannon coding is a basic prefix free coding. There is some confusion in literature about Shannon, Fano and
+Shannon-Fano codes, e.g. see Naming section of https://en.wikipedia.org/wiki/Shannon–Fano_coding.
```
shubhamchandak94, line 3: somewhat paranoid - but maybe better to put the permalink to wikipedia page - left pane of wikipedia page has a link

compressors/fano_coder.py
```diff
@@ -0,0 +1,160 @@
...
+        self.sorted_prob_dist = ProbabilityDist.get_sorted_prob_dist(prob_dist.prob_dict)
+        # initialize root node of Fano Tree
+        self.root_node = BinaryNode(id=None)
+
+        # build tree returns the root node
+        super().__init__(root_node=self.build_fano_tree(self.root_node, self.sorted_prob_dist))
+
+
+    @staticmethod
+    def criterion(dict_item):
```
shubhamchandak94, line 27: I think this should be a local function defined inside `split_prob_dist_into_two` since it is not used anywhere else

compressors/fano_coder.py
```diff
@@ -0,0 +1,160 @@
...
+            'right': {}
+        }
+
+        # all alphabets which are nearest to 0.5 cumulative probability in sorted list are put to left tree and rest
+        # to right, such that we have a most balanced probabilistic tree subsets.
+        min_diff_code = min(cumulative_prob_dict.items(), key=FanoTree.criterion)[0]
+        curr_dict = recursive_prob_dicts['left']
+        for s, cum_prob in cumulative_prob_dict.items():
+            if s == min_diff_code:
+                curr_dict = recursive_prob_dicts['right']
```
shubhamchandak94, line 57: add comment - `# after min_diff_code switch to right tree for putting the symbols`

compressors/fano_coder.py
```diff
@@ -0,0 +1,160 @@
...
+        # Split the tree into left and right nodes based on cumulative probability and recursively build
+        # FanoTree for these trees
+        recursive_prob_dicts = {
+            'left': {},
+            'right': {}
+        }
+
+        # all alphabets which are nearest to 0.5 cumulative probability in sorted list are put to left tree and rest
+        # to right, such that we have a most balanced probabilistic tree subsets.
+        min_diff_code = min(cumulative_prob_dict.items(), key=FanoTree.criterion)[0]
```
shubhamchandak94, line 53: I don't understand the variable name `min_diff_code` (the `code` part specifically) - should it be called `min_diff_symbol`?

compressors/fano_coder.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        Given a normalized and sorted probability distribution, split it into two sets of approximately equal
+        probabilities as described in Fano's algorithm.
+        """
+        # Get cumulative probability dict
+        cumulative_prob_dict = norm_sort_prob_dist.cumulative_prob_dict
+
+        # Split the tree into left and right nodes based on cumulative probability and recursively build
+        # FanoTree for these trees
+        recursive_prob_dicts = {
```
shubhamchandak94, line 46: not convinced about naming this recursive, because we don't do any recursion in the split function. One suggestion is `prob_dicts_after_split` or something similar

compressors/fano_coder.py
```diff
@@ -0,0 +1,160 @@
...
+from typing import Any, Tuple
+from utils.bitarray_utils import BitArray
+from utils.test_utils import get_random_data_block, try_lossless_compression
+from compressors.prefix_free_compressors import PrefixFreeTree, PrefixFreeEncoder, PrefixFreeDecoder, BinaryNode
+from core.prob_dist import ProbabilityDist
+
+
+class FanoTree(PrefixFreeTree):
+    def __init__(self, prob_dist):
+        super().__init__(prob_dist)
```
shubhamchandak94, line 16: sorry, what's this doing?

shubhamchandak94: I see - I think I understand what's happening. Is doing stuff to super like this considered good practice even when we don't know what super is? Perhaps better is to just set to self here and have the super access these like `self.tree.prob_dist`?

shubhamchandak94: More generally, I think for any fancy python features we use - let's add a link to docs so people can read up. I am sure there are others like me who haven't had formal Python training struggle with some of the advanced concepts. I don't think that should come in the way of someone focused on understanding compression concepts.

compressors/fano_coder.py
```diff
@@ -0,0 +1,160 @@
...
+        # most balanced
+        recursive_prob_dicts = FanoTree.split_prob_dist_into_two(norm_sort_prob_dist)
+
+        # Call recursion
+        # if only 1 symbol in either left or right tree, just assign it as a child and we don't have to call recursion
+        for branch, curr_dict in recursive_prob_dicts.items():
+            norm_prob_dict = ProbabilityDist.normalize_prob_dict(curr_dict)
+
+            # if pointer is None, the object is not updated
+            # https://stackoverflow.com/questions/55777748/updating-none-value-does-not-reflect-in-the-object
```
shubhamchandak94, line 76: move this outside the loop, and add comment like 
```
# we initialize the right and left child here and later separately update/recurse on them in the loop below (the loop has just 2 steps: left and right. Initialization is important since if we leave these as None, then a pattern like var = root_node.left_child; var.id = new_id` won't work because `var` would be just `None` and not a pointer
```

compressors/huffman_coder.py
```diff
@@ -85,14 +85,15 @@ class HuffmanEncoder(PrefixFreeEncoder):
     """
     PrefixFreeEncoder already has a encode_block function to encode the symbols once we define a encode_symbol function
     for the particular compressor.
-    PrefixFreeTree provides encode_symbol given a PrefixFreeTree
+    PrefixFreeTree provides get_encoding_table given a PrefixFreeTree
```
shubhamchandak94, line 88: should it be `PrefixFreeEncoder provides get_encoding_table given a PrefixFreeTree`

compressors/shannon_coder.py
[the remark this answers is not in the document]
shubhamchandak94: Add link to wiki page somewhere in the code?

compressors/shannon_coder.py
```diff
@@ -0,0 +1,156 @@
+"""Shannon Coding
+Shannon coding is a basic prefix free code. There is some confusion in literature about Shannon, Fano and
+Shannon-Fano codes, e.g. see Naming section of https://en.wikipedia.org/wiki/Shannon–Fano_coding.
```
shubhamchandak94, line 3: same comment as in Fano code about permalinks

compressors/shannon_coder.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+
+    def __init__(self, prob_dist: ProbabilityDist):
+        self.prob_dist = prob_dist
+        # sort the probability distribution in decreasing probability and get cumulative probability which will be
+        # used for encoding
+        self.sorted_prob_dist = ProbabilityDist.get_sorted_prob_dist(prob_dist.prob_dict)
+        self.cum_prob_dict = self.sorted_prob_dist.cumulative_prob_dict
+        # construct the tree and set the root_node of PrefixFreeTree base class
+        super().__init__(root_node=self.build_shannon_tree())
```
shubhamchandak94, line 27: same comment as Fano on `super().__init__()` stuff

compressors/shannon_coder.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def encode_alphabet(self, symbol) -> BitArray:
```
shubhamchandak94, line 29: why not just call it `encode_symbol` maybe with underscore if we want to distinguish from the prefix coder version?

compressors/shannon_coder.py
```diff
@@ -0,0 +1,156 @@
...
+        Returns:
+            the pointer to root node of the tree so far
+        """
+        # initialize the curr_node, code_so_far temporary var
+        curr_node = root_node
+        code_so_far = BitArray()
+        code_len = len(code)
+
+        for i, bit in enumerate(code):
+            # if pointer is None, the object is not updated
```
shubhamchandak94, line 57: add 
```
i.e., curr_node.right_child being None would not allow us to keep the right child in a variable for updating further.
```

compressors/shannon_coder.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # compute encode length
+        encode_len = math.ceil(self.prob_dist.log_probability(symbol))
+
+        # the encode is the binary representation of the cumulative probability in ascending order,
+        _, code = float_to_bitarrays(cum_prob, encode_len)
+        return code
+
+    @staticmethod
+    def _build_tree_from_code(symbol, code, root_node) -> BinaryNode:
```
shubhamchandak94, line 41: I think this should sit outside Shannon code, maybe in the Prefix Tree code. Also I feel a better interface is to have:
```
    def build_shannon_tree(self):
        """
        For all symbols in the alphabet, get its code, and add it to the PrefixFreeTree.
        """
        root_node = BinaryNode(id=None)
        code = self.generate_codewords(s) # this internally uses encode_alphabet in a loop, returns dict from symbol to codeword
        root_node = self._build_tree_from_code(code)
        return root_node
```
where the `_build_tree_from_code` should take the entire code as a dict and generate the whole tree at once.

shubhamchandak94: Also wonder if it's natural to detect non-prefix codes as part of the `_build_tree_from_code` function

core/prob_dist.py
```diff
@@ -32,6 +32,24 @@ def alphabet(self):
     def prob_list(self):
         return [self.prob_dict[s] for s in self.alphabet]
 
+    @classmethod
+    def get_sorted_prob_dist(cls, prob_dict, reverse=True):
```
shubhamchandak94, line 36: I don't like making `reverse=True` default - we should follow the standard interface to sort in Python and have it as False by default

core/prob_dist.py
```diff
@@ -32,6 +32,24 @@ def alphabet(self):
...
+        return cls(dict(sorted(prob_dict.items(), key=lambda x: x[1], reverse=reverse)))
+
+    @classmethod
+    def normalize_prob_dict(cls, prob_dict):
+        """
+        normalizes dict -> dict_norm so that the sum of values is 1
+        wraps dict_norm as a ProbabilityDist
+        """
+        sum_p = sum(prob_dict.values())
+        return cls(dict([[a, b / sum_p] for a, b in prob_dict.items()]))
```
shubhamchandak94, line 51: can also write `cls({a: b/sum_p for a, b in prob_dict.items()})`

core/prob_dist.py
```diff
@@ -107,6 +125,20 @@ def test_prob_creation_and_validation(self):
...
         _ = ProbabilityDist(dist)
 
+    def test_sorted_prob_dist(self):
+        """
+        checks if sorting works as expected
+        """
+        alphabet = list(range(10))
+        dist = {i: (i + 1) / 55 for i in alphabet}
+
+        sorted_PD = ProbabilityDist.get_sorted_prob_dist(dist)
```
shubhamchandak94, line 135: also add equality check to make sure the dicts are same except for reordering

compressors/shannon_coder.py
shubhamchandak94: On further thought, if we are having this trouble again and again, maybe we want to make the __init__ behavior of a node to instantiate the left and right child?

shubhamchandak94 (commented):
Thanks Pulkit! Sorry I took so long for the review. I have left some comments now, hopefully they are helpful.

compressors/fano_coder.py
kedartatwawadi: +1 yes, discussed the same with @tpulkit last week.

compressors/fano_coder.py
kedartatwawadi: +1, had discussed with pulkit to just have left, right.. as it makes things more clear ..

compressors/fano_coder.py
tpulkit: super is a standard way to initialize variable of a base class in an inherited class. We know in this case base class requires prob_dist which our inherited class also requires. We can add a comment to the doc: https://stackoverflow.com/questions/576169/understanding-python-super-with-init-methods

Would this work?

compressors/fano_coder.py
tpulkit: thanks for the detailed comment.

compressors/huffman_coder.py
tpulkit: yes, thanks!

compressors/shannon_coder.py
tpulkit: it's encoding an alphabet in the dictionary not a symbol. I wanted to be clear wrt alphabet and symbol. It will only be used once when the original alphabet is defined. I have modified it to` _encode_alphabet`. Lmk if u have any other concerns.

compressors/shannon_coder.py
tpulkit: I have updated the comment but in general I like your suggestion. @kedartatwawadi: thoughts?

compressors/shannon_coder.py
tpulkit: thanks for the suggestion. Updated. 
Not sure about detecting comment: ideally there should be a detection during the creation of PrefixFreeClass and if at any point it becomes non-prefix free it should throw an error?

core/prob_dist.py
tpulkit: Sorry I don't understand this? Why would you do that if we want in reverse sorted order and we are using a built-in function?

compressors/fano_coder.py
kedartatwawadi: Yeah, I think super() is a reasonably commonly used feature. @shubhamchandak94: For single inheritance, super() is just the parent class.. I believe we have this quite a bit in our codebase already

core/prob_dist.py
kedartatwawadi: @tpulkit: I kind of agree with shubham. Also, the function name`get_sorted_prob_dist` intuitively implies sorting in increasing order.. so maybe we should keep that as the default.

compressors/shannon_coder.py
tpulkit: Also move the whole function of generating tree from codewords to PrefixFreeCompressor

compressors/shannon_coder.py
tpulkit: Raised issue #43

core/prob_dist.py
tpulkit: make default parameter False, and then call appropriately

shubhamchandak94 (approved):
Looks good! Thanks for making the changes :)

kedartatwawadi (approved):
LG! yay!!

PR #39, opened 2022-04-09
Cleanup and simplify PrefixFreeTree and encoders

kedartatwawadi:
Based on feedback from @tpulkit: I was thinking of simplifying the `PrefixFreeTree` etc. 
The issue earlier was that there were too many sublcasses, which were making it difficult to understand the code.

Some simplifications: 
- [x] Removed `PrefixFreeNode`.. as it added very less value. (We had added the field "code".. but we really don't need a new field). 
- [x] Moved the function `get_encoding_table` to `PrefixFreeTree` .. as it is really a function of the tree. 
- [x] All prefix-free coders can now only subclass `PrefixFreeEncoder` and `PrefixFreeDecoder`. NO `PrefixFreeTreeEncoder` etc. .. they can however construct define their own tree internally. I thought this increases the transparency, and makes the implemented coders easier to understand. 

Feedback welcome! @tpulkit: If we merge this, you might have to change a few things on your end. It will probably make your life much easier.

### DIFF
--- a/compressors/huffman_coder.py
+++ b/compressors/huffman_coder.py
@@ -45,7 +45,7 @@ class HuffmanTree(PrefixFreeTree):
         """
         # Lets say we have symbols {1,2,3,4,5,6} with prob {p1, p2,...p6}
         # We first start by initializing a list
-        # [ HuffmanNode(id=3, prob=p3), (HuffmanTree(id=6, prob=p6),  ]
+        # [..., HuffmanNode(id=3, prob=p3), (HuffmanNode(id=6, prob=p6),  ...]
 
         node_list = []
         for a in self.prob_dist.alphabet:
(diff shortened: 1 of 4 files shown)


shubhamchandak94 (approved):
Looks good to me. I have gone over your PR description and the new code and don't really have any issues with it. Happy to provide feedback on any specific part if relevant.

tpulkit:
I really like this new structure! Definitely simplifies a lot of confusing things in the last version.
I think comments can be better. I will just make those changes directly in the branch.

tpulkit:
I mainly added better comments but also simplified encoding aspect of PrefixFreeTree by providing encode_symbol function (similar to decode symbol function) as part of the tree.

PR #40, opened 2022-04-11
Implement Frequencies + other utility functions

kedartatwawadi:
I didn't get time to write about arithmetic coding, so that PR is still in draft version (i.e. not ready to review) 
I think it might be beneficial to merge some of the utils until then.


core/prob_dist.py
```diff
@@ -100,3 +100,93 @@ def test_prob_creation_and_validation(self):
 
         # check if this works
         _ = ProbabilityDist(dist)
+
+
+def get_mean_log_prob(prob_dist: ProbabilityDist, data_block) -> float:
+    """computes the average log_probability of the input data_block given the probability distribution
+    prob_dist. This acts as a fundamental limit on what any compressor designed for distribution
```
shubhamchandak94, line 107: "This acts as a fundamental limit on what any ..." - I would make this weaker statement. It's intuitively correct for most standard entropy coders "designed for distribution prob_dist", but one can always created weird codes that give small codewords to low probability symbols to beat this bound for certain sequences

core/prob_dist.py
```diff
@@ -100,3 +100,93 @@ def test_prob_creation_and_validation(self):
...
+
+class Frequencies:
+    """
+    Wrapper around a frequency dict
+    NOTE: Frequencies is a typical way to represent probability distributions using integers
+    """
+
+    def __init__(self, freq_dict=None):
+
+        # NOTE: We use the fact that since python 3.6, dictionaries in python are
```
shubhamchandak94, line 131: where do we use this fact? Is it for the lists?

core/prob_dist.py
```diff
@@ -100,3 +100,93 @@ def test_prob_creation_and_validation(self):
...
+
+    @property
+    def alphabet(self):
+        return list(self.freq_dict)
+
+    @property
+    def freq_list(self):
+        return [self.freq_dict[s] for s in self.alphabet]
+
+    @property
```
shubhamchandak94, line 150: TODO - do the cached thing

core/prob_dist.py
shubhamchandak94: I guess also for cumulative

utils/bitarray_utils.py
```diff
@@ -4,8 +4,16 @@
 from typing import Tuple
 
 
-def get_bit_width(size) -> int:
-    return int(np.ceil(np.log2(size)))
+def get_bit_width(x) -> int:
```
shubhamchandak94, line 7: TODO - add test for this, checking for boundary values like powers of 2, powers of 2 minus 1 (looks correct to me but I have made mistakes in the past, so better to add simple tests)
also add assertion for positivity
do we want to support x=0

utils/bitarray_utils.py
```diff
@@ -18,7 +26,7 @@ def uint_to_bitarray(x: int, bit_width=None) -> BitArray:
     converts an unsigned into to bits.
     if bit_width is provided then data is converted accordingly
     """
-    return int2ba(x, length=bit_width)
+    return int2ba(int(x), length=bit_width)
```
shubhamchandak94, line 29: why converting to int? Is it better to give an error than to silently convert to int

utils/test_utils.py
```diff
@@ -69,7 +69,8 @@ def try_lossless_compression(
     output_bits_block = encoder.encode_block(data_block)
 
     # test decode
-    decoded_block, _ = decoder.decode_block(output_bits_block)
+    decoded_block, num_bits_consumed = decoder.decode_block(output_bits_block)
+    assert num_bits_consumed == len(output_bits_block)
```
shubhamchandak94, line 73: Nice. Wonder how we missed this before

core/prob_dist.py
kedartatwawadi: rewrote as 
```
NOTE: although not a funcdamental limit, this is roughly what a compressor designed for the given distribution prob_dist will aim to achieve"
```

core/prob_dist.py
```diff
@@ -100,3 +100,93 @@ def test_prob_creation_and_validation(self):
...
+        prob_dist (ProbabilityDist): specified probability distribution used to compute log_probability
+        data_block (DataBlock): input for which avg log probability needs to be computed
+    """
+
+    # get avg log probability for the input
+    log_prob = 0
+    for s in data_block.data_list:
+        log_prob += prob_dist.log_probability(s)
+    avg_log_prob = log_prob / data_block.size
+    return avg_log_prob
```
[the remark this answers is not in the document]
kedartatwawadi: done

core/prob_dist.py
kedartatwawadi: yes exactly

core/prob_dist.py
kedartatwawadi: done

utils/bitarray_utils.py
kedartatwawadi: good point, I think there is indeed a mistake here! 
if x = 4, this returns 2, which is not enough :-|

kedartatwawadi: fixed + added tests

utils/bitarray_utils.py
kedartatwawadi: yeah, not needed it looks like

kedartatwawadi: added an assert

utils/test_utils.py
kedartatwawadi: actually want to improve this test..  by optionally adding some extra bits to encoder output. This is to simulate the scenario where there are additional bits beyond the bitstream encoded. Some compressors (like range coder, AEC) support this .

kedartatwawadi: @shubhamchandak94: take a re-look

shubhamchandak94 (approved):

core/prob_dist.py
tpulkit: I don't see the change in the latest committed files.

core/prob_dist.py
tpulkit: Why do you need it to be OrderedDict? We don't really care order in which symbols were added to the dict, correct?

core/prob_dist.py
kedartatwawadi: I did update it a bit, but still kept the "fundamental limit" business.. ll update

core/prob_dist.py
kedartatwawadi: We do need ordering. for example when we return the `freq_list` or the `cumulative_freq_list`.. we assume that the first frequency in the list corresponds to  the first symbol (for example).

shubhamchandak94: I think the important part in ordering is the consistency. The particular ordering according to insertion order is probably not that important because we will almost always insert all symbols at the beginning and later only update frequencies

tpulkit: Got it. Thanks.

core/prob_dist.py
tpulkit: English can be improved, but I will merge it for now. I think I will have a look at all the comments before class just to see the language makes sense. I think many of my comments are also not good grammatically.

tpulkit (commented):
Looks good. Merging.

tpulkit (approved):

PR #41, opened 2022-04-13
rANS 

kedartatwawadi:
I added decent amount of comments, but let me know if something is not clear as you review the code

### DIFF
--- a/compressors/rANS.py
+++ b/compressors/rANS.py
@@ -67,7 +67,7 @@ from utils.bitarray_utils import BitArray, get_bit_width, uint_to_bitarray, bita
 from core.data_block import DataBlock
 from core.prob_dist import Frequencies, get_mean_log_prob
 from utils.test_utils import get_random_data_block, try_lossless_compression
-from utils.misc_utils import cache, Symbol
+from utils.misc_utils import cache
 
 
 @dataclass
@@ -113,7 +113,7 @@ class rANSParams:
 
 class rANSEncoder(DataEncoder):
     """rANS Encoder
-    
+
     Detailed information in the overview
     """
 
@@ -127,7 +127,7 @@ class rANSEncoder(DataEncoder):
         self.freqs = freqs
         self.params = rans_params
 
-    def rans_base_encode_step(self, s: Symbol, state: int):
+    def rans_base_encode_step(self, s, state: int):
         """base rANS encode step
 
         updates the state based on the input symbols s, and returns the updated state
@@ -141,21 +141,13 @@ class rANSEncoder(DataEncoder):
     @cache
     def max_state_val(self, symbol):
         """
-        TODO: 
+        max value the state can be before calling rans_base_encode_step function
         """
         f = self.freqs.frequency(symbol)
         return self.params.RANGE_FACTOR * f * (1 << self.params.NUM_BITS_OUT) - 1
 
     def shrink_state(self, state: int, next_symbol) -> Tuple[int, BitArray]:
-        """TODO
-
-        Args:
-            state (int): _description_
-            next_symbol (_type_): _description_
-
-        Returns:
-            Tuple[int, BitArray]: _description_
-        """
+        """stream out the lower bits of the state, until the state is below self.max_state_val(next_symbol)"""
         out_bits = BitArray("")
 
         # output bits to the stream to bring the state in the range for the next encoding
@@ -176,7 +168,7 @@ class rANSEncoder(DataEncoder):
             state (int): the rANS state
 
         Returns:
-            state (int), symbol_bitarray (BitArray): 
+            state (int), symbol_bitarray (BitArray):
         """
         # output bits to the stream so that the state is in the acceptable range
         # [L, H] *after*the `rans_base_encode_step`
--- a/compressors/rANS.py
+++ b/compressors/rANS.py
@@ -1,5 +1,8 @@
 """Streaming rANS (range Asymmetric Numeral Systems) implementation
 
+NOTE: Detailed algorithm description and discussion is on the wiki page:
+https://github.com/kedartatwawadi/stanford_compression_library/wiki/Asymmetric-Numeral-Systems
+
 ## Core idea
 - the theoretical rANS Encoder maintains an integer `state` 
 - For each symbol s, the state is updated by calling: 
--- a/compressors/rANS.py
+++ b/compressors/rANS.py
@@ -1,3 +1,64 @@
+"""Streaming rANS (range Asymmetric Numeral Systems) implementation
+
+## Core idea
+- the theoretical rANS Encoder maintains an integer `state` 
+- For each symbol s, the state is updated by calling: 
+    ```python
+    # encode step
+    state = rans_base_encode_step(s, state)
+    ```
+    the decoder does the reverse by decoding the s and retrieving the prev state
+    ```python
+    # decode step
+    s, state = rans_base_decode_step(state)
+    ```
+- In the theoretical rANS version, the state keeps on increasing every time we call `rans_base_encode_step`
+  To make this practical, the rANS encoder ensures that after each encode step, the `state` lies in the acceptable range
+  `[L, H]`, where `L,H` are predefined interval values.
+
+  ```
+  state lies in [L, H], after every encode step
+  ```
+  To ensure this happens the encoder shrinks the `state` by streaming out its lower bits, *before* encoding each symbol. 
+  This logic is implemented in the function `shrink_state`. Thus, the full encoding step for one symbol is as follows:
+
+  ```python
+    ## Encoding one symbol
+    # output bits to the stream to bring the state in the range for the next encoding
+        state, out_bits = self.shrink_state(state, s)
+        encoded_bitarray = out_bits + encoded_bitarray
+
+    # core encoding step
+    state = self.rans_base_encode_step(s, state)
+  ```
+
+  The decoder does the reverse operation of `expand_state` where, after decoding a symbol, it reads in the a few bits to
+  re-map and expand the state to lie within the acceptable range [L, H]
+  Note that `shrink_state` and `expand_state` are inverses of each other
+
+  Thus, the  full decoding step 
+  ```python
+    # base rANS decoding step
+    s, state = self.rans_base_decode_step(state)
+
+    # remap the state into the acceptable range
+    state, num_bits_used_by_expand_state = self.expand_state(state, encoded_bitarray)
+  ```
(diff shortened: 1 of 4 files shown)


utils/bitarray_utils.py
```diff
@@ -29,8 +29,8 @@ def uint_to_bitarray(x: int, bit_width=None) -> BitArray:
     converts an unsigned into to bits.
     if bit_width is provided then data is converted accordingly
     """
-    assert isinstance(x, int)
-    return int2ba(x, length=bit_width)
+    assert isinstance(x, (int, np.integer))
```
tpulkit, line 32: Does this check specifically that x should be numpy integer?

shubhamchandak94:
In the wiki (streaming rANS section):
```
M*t = rans_base_encode_step(freq[s]*t,s)
2*M*t = rans_base_encode_step(2*freq[s]*t,s)
```
I am a bit confused here - should there be a `cumul[s]` term as well?

compressors/rANS.py
```diff
@@ -0,0 +1,335 @@
...
+    # base rANS decoding step
+    s, state = self.rans_base_decode_step(state)
+
+    # remap the state into the acceptable range
+    state, num_bits_used_by_expand_state = self.expand_state(state, encoded_bitarray)
+  ```
+- For completeness: the acceptable range `[L, H]` are given by:
+  L = RANGE_FACTOR*total_freq
+  H = (2**NUM_BITS_OUT)*RANGE_FACTOR*total_freq - 1)
+  Why specifically these values? Look at the *Streaming-rANS* section on https://kedartatwawadi.github.io/post--ANS/
```
shubhamchandak94, line 50: also add link to wiki somewhere in the code?

compressors/rANS.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+## References
+1. Original Asymmetric Numeral Systems paper:  https://arxiv.org/abs/0902.0271
+2. Interactive write-up on rANS: https://kedartatwawadi.github.io/post--ANS/
+
+## TODO
+1. Implement alias-coding style rANS
```
shubhamchandak94, line 58: what's this (in short)?

compressors/rANS.py
```diff
@@ -0,0 +1,335 @@
...
+    More details in the overview
+    """
+
+    # num bits used to represent the data_block size
+    DATA_BLOCK_SIZE_BITS: int = 32
+
+    # the encoder can output NUM_BITS_OUT at a time when it performs the state shrinking operation
+    NUM_BITS_OUT: int = 1  # number of bits
+
+    # rANS state is limited to the range [RANGE_FACTOR*total_freq, (2**NUM_BITS_OUT)*RANGE_FACTOR*total_freq - 1)]
```
shubhamchandak94, line 85: What is the tradeoff associated with RANGE_FACTOR? I can understand that it can't be too large because H is at most 32 bit (precision). On that note, is there a limit here on total_freq so we don't have overflow when doing `(2**NUM_BITS_OUT)*RANGE_FACTOR*total_freq` (python doesn't have int overflow but I'm talking more generally for a finite precision implementation)

shubhamchandak94: Does setting `NUM_BITS_OUT` to 8 lead to a range coding equivalent? Is that something people use?

shubhamchandak94: What is the importance of choosing `RANGE_FACTOR` (`t` from wiki) as something large rather than just 1. Is there precision loss on choosing it too small?

compressors/rANS.py
```diff
@@ -0,0 +1,335 @@
...
+    # RANGE_FACTOR is a base parameter controlling this range
+    RANGE_FACTOR_BITS: int = 16
+    RANGE_FACTOR: int = 1 << RANGE_FACTOR_BITS
+
+    def initial_state(self, freqs: Frequencies) -> int:
+        """the initial state from which rANS encoding begins
+
+        NOTE: the choice of  this state is somewhat arbitrary, the only condition being, it should lie in the acceptable range
```
shubhamchandak94, line 93: for arbitrary choices, did we follow some standard reference implementation? Also does it make sense to define L and H as parameters for clarity?

compressors/rANS.py
```diff
@@ -0,0 +1,335 @@
...
+        # initialize the output
+        encoded_bitarray = BitArray("")
+
+        # initialize the state
+        state = self.params.initial_state(self.freqs)
+
+        # update the state
+        for s in data_block.data_list:
+            state, symbol_bitarray = self.encode_symbol(s, state)
+            encoded_bitarray = symbol_bitarray + encoded_bitarray
```
shubhamchandak94, line 193: This prepending - is this the standard method or did we use it for its simplicity (as opposed to decoding in reverse or something else)?

shubhamchandak94: Prepending just seems very anti streaming and inefficient since it involves lots of memcopies and allocations (unless bitarray uses a deque internally)

compressors/rANS.py
```diff
@@ -0,0 +1,335 @@
...
+
+        Args:
+            cumulative_freqs_list (List): the sorted list of cumulative frequencies
+                For example: freqs_list = [2,7,3], cumulative_freqs_list [0,2,9]
+            slot (int): the value to search in the sorted list
+
+        Returns:
+            bin: the bin in which the slot lies
+        """
+        bin = np.searchsorted(cumulative_freqs_list, slot, side="right") - 1
```
shubhamchandak94, line 229: add comment on why side="right" is important, maybe as part of example in doc comment

compressors/rANS.py
shubhamchandak94: I understand that this is just L

compressors/rANS.py
```diff
@@ -0,0 +1,335 @@
...
+        # Finally, as a sanity check, ensure that the end state should be equal to the initial state
+        assert state == self.params.initial_state(self.freqs)
+
+        return DataBlock(decoded_data_list), num_bits_consumed
+
+
+######################################## TESTS ##########################################
+
+
+def _test_rANS_coding(freq, data_size, seed):
```
shubhamchandak94, line 298: we have been getting a lot of these functions recently - can think of abstracting it out (not necessary as part of this PR)

compressors/rANS.py
```diff
@@ -0,0 +1,335 @@
...
+    prob_dist = freq.get_prob_dist()
+
+    # generate random data
+    data_block = get_random_data_block(prob_dist, data_size, seed=seed)
+
+    # get optimal codelen
+    avg_log_prob = get_mean_log_prob(prob_dist, data_block)
+
+    # create encoder decoder
+    rans_params = rANSParams()
```
shubhamchandak94, line 309: also test a couple different values of NUM_BITS_OUT and RANGE_FACTOR?

utils/bitarray_utils.py
shubhamchandak94: To Pulkit's question:
From https://www.w3schools.com/python/ref_func_isinstance.asp
> If the type parameter is a tuple, this function will return True if the object is one of the types in the tuple.

shubhamchandak94: @kedartatwawadi - are we using numpy integer types anywhere?

shubhamchandak94 (commented):
Thanks Kedar! Generally looks great after understanding the details from the wiki. I have left some minor comments here and there.

compressors/rANS.py
kedartatwawadi: ll do!

compressors/rANS.py
kedartatwawadi: I will briefly mention this tomorrow.. but as a hint look at the puzzle I posted on slack!

compressors/rANS.py
kedartatwawadi: in my understanding the main benefit of choosing `t` to be large is that the division during the encoding `x//freq[s]` is more precise. There is no other benefit I think

kedartatwawadi: Regarding `NUM_BITS_OUT` yes, people use 8bit, 16bit and 32bit "range" coder variants. that is precisely where the name `rANS` comes from. 

See: https://github.com/rygorous/ryg_rans

kedartatwawadi: image

compressors/rANS.py
kedartatwawadi: Yeah, now that i think of it, L, H might be better to be defined beforehand. Also, what do you think about including `freq` in the rANSParams`? that would simplify quite a few things

compressors/rANS.py
kedartatwawadi: yes, practically encoding is performed from end to the beginning

compressors/rANS.py
shubhamchandak94: including `freq` in the rANSParams makes sense

compressors/rANS.py
tpulkit, line 298: also please add a test of just the encoder -- checking if encoder is outputting expected codewords as we would expect theoretically.

kedartatwawadi: thanks for reminding! ll do

compressors/rANS.py
kedartatwawadi: done

compressors/rANS.py
kedartatwawadi: yes ll do that in a followup PR

compressors/rANS.py
kedartatwawadi: done

compressors/rANS.py
kedartatwawadi: I was concerned about it too.. practically it seems a large memory is allocated beforehand, and then it is just filled backwards.. https://github.com/rygorous/ryg_rans/blob/c9d162d996fd600315af9ae8eb89d832576cb32d/main.cpp#L176

unfortunately there is no way around this (we have to either decode from behind or encode from behind)

utils/bitarray_utils.py
kedartatwawadi: resolved during our call

compressors/rANS.py
kedartatwawadi: not doing this now.. ll refactor if and when we implement the adaptive version

compressors/rANS.py
kedartatwawadi: Added a comment

shubhamchandak94 (approved):
Looks good!

compressors/rANS.py
kedartatwawadi: done @tpulkit

PR #44, opened 2022-05-15
Table ANS implementation

kedartatwawadi:
- Some diffs are because I ran `black . --line-length 100`.. not sure why the tests didnt catch this when we merged the PR.
- Cleaned up `rANS`.. I thought it might be simpler to adds `freqs` to list of params.. when we implement adaptive rANS (if and when!) we can separate things out.
- Implented `tANS`.. added a description to wiki.. it might have some typos.. also I need to add examples + figures at some point


compressors/tANS.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def __post_init__(self):
+        super().__post_init__()
+        ## restrict some param values for tANS
+        # to make the rANS cachable, the total_freq needs to be a power of 2
+        assert is_power_of_two(
+            self.freqs.total_freq
+        ), "Please normalize self.freqs.total_freq to be a power of two"
+
+        # NOTE: NUM_BITS_OUT != 1, probably doesn't make practical sense for tANS?
```
shubhamchandak94, line 41: Can you elaborate why a similar technique as rANS can't work for tANS for NUM_BITS_OUT>0?
Also, what's the typical practice (e.g., in FSE library)?

compressors/rANS.py
```diff
@@ -380,29 +344,48 @@ def test_check_encoded_bitarray():
...
-        rANSParams(RANGE_FACTOR_BITS=12),
+    params_list = [
+        rANSParams(freqs_list[0]),
+        rANSParams(freqs_list[1]),
+        rANSParams(freqs_list[2], NUM_BITS_OUT=8),
+        rANSParams(freqs_list[3], RANGE_FACTOR=1 << 12),
+        rANSParams(freqs_list[4], RANGE_FACTOR=1 << 4),
     ]
-    for freq, param in zip(freqs, params):
-        _test_rANS_coding(freq, param, DATA_SIZE, seed=0)
```
shubhamchandak94, line 408: what happened here? Why did we move away from the function interface?

shubhamchandak94 (commented):
left a couple comments. Will take another look after I read the wiki again (have forgotten some of the rANS and tANS concepts)

shubhamchandak94 (approved):
Looks good

compressors/rANS.py
```diff
@@ -84,31 +87,28 @@ class rANSParams:
 
     # rANS state is limited to the range [RANGE_FACTOR*total_freq, (2**NUM_BITS_OUT)*RANGE_FACTOR*total_freq - 1)]
     # RANGE_FACTOR is a base parameter controlling this range
-    RANGE_FACTOR_BITS: int = 16
-    RANGE_FACTOR: int = 1 << RANGE_FACTOR_BITS
-
-    def initial_state(self, freqs: Frequencies) -> int:
-        """the initial state from which rANS encoding begins
+    RANGE_FACTOR: int = 1 << 16
```
tpulkit, line 90: The write-up doesn't mention RANGE_FACTOR. Maybe add here that RANGE_FACTOR is same as M in write-up, or if this is a more commonly used term, we should add it in write-up.

tpulkit: also, I thought you would choose NUM_BITS_OUT to be 8 or 16 for speed reasons. Is it a reason it is 1 here. Should we test that the code works with higher NUM_BITS_OUT as well and leads to some speedup depending on architecture?

compressors/rANS.py
```diff
@@ -117,41 +117,32 @@ class rANSEncoder(DataEncoder):
...
-        return self.params.RANGE_FACTOR * f * (1 << self.params.NUM_BITS_OUT) - 1
-
     def shrink_state(self, state: int, next_symbol) -> Tuple[int, BitArray]:
-        """stream out the lower bits of the state, until the state is below self.max_state_val(next_symbol)"""
+        """stream out the lower bits of the state, until the state is below params.max_shrunk_state[next_symbol]"""
         out_bits = BitArray("")
 
         # output bits to the stream to bring the state in the range for the next encoding
-        while state > self.max_shrunk_state_val(next_symbol):
+        while state > self.params.max_shrunk_state[next_symbol]:
```
tpulkit, line 145: Is there a reason you don't store `(1 << self.params.NUM_BITS_OUT)` as a __post_init__ parameter in rANSParams instead of recalculating it each time?

compressors/rANS.py
```diff
@@ -117,41 +117,32 @@ class rANSEncoder(DataEncoder):
...
             _bits = uint_to_bitarray(
                 state % (1 << self.params.NUM_BITS_OUT), bit_width=self.params.NUM_BITS_OUT
             )
```
tpulkit, line 148: `out_bits = _bits + out_bits`: is prepending, write-up says appending during putting bits out.

compressors/rANS.py
```diff
@@ -295,43 +283,19 @@ def decode_block(self, encoded_bitarray: BitArray):
...
-        data_block, encoder, decoder, add_extra_bits_to_encoder_output=True
-    )
-
-    # avg codelen ignoring the bits used to signal num data elements
-    avg_codelen = encode_len / data_block.size
-    print(f"rANS coding: Optical codelen={avg_log_prob:.3f}, rANS codelen: {avg_codelen:.3f}")
-    assert is_lossless
-
-
 def test_check_encoded_bitarray():
```
tpulkit, line 294: In the comment: 
`    # NOTE: the encoded_bitstream looks like = [<data_size_bits>, <final_state_bits>,<s0_bits>, <s1_bits>, ..., <s3_bits>]
`

Shouldn't it be:
`    # NOTE: the encoded_bitstream looks like = [<data_size_bits>, <final_state_bits>,<s_n-1_bits>, <s_n-2_bits>, ..., <s0_bits>]
`
I am assuming s_0 is the state you get after encoding first symbol. in this case after encoding A. You are prepending the states. Code makes sense, just trying to see if I understand consistently wrt comments.

compressors/tANS.py
```diff
@@ -0,0 +1,380 @@
+"""tANS v1 (table ANS) implementation 
+
+NOTE: tANS v1 is ina  way cached rANS implementation. There are other variants of tANS possible
```
tpulkit, line 3: is in a way

compressors/tANS.py
tpulkit: Sorry, I don't understand this. Why would it be exponential in `NUM_BITS_OUT`? The table size shouldn't depend on how many bits I am sending to ensure the range remains in [L,H]. It should depend on the parameters once the state is in that range. Am I missing something?

compressors/tANS.py
```diff
@@ -0,0 +1,380 @@
...
+        # output bits to the stream so that the state is in the acceptable range
+        # [L, H] *after*the `rans_base_encode_step`
+        symbol_bitarray = BitArray("")
+
+        # shrink state x before calling base encode
+        num_out_bits = self.shrink_state_num_out_bits_base_table[s]
+        if state >= self.shrink_state_thresh_table[s]:
+            num_out_bits += 1
+
+        out_bits = uint_to_bitarray(state)[-num_out_bits:] if num_out_bits else BitArray("")
```
tpulkit, line 141: basic python question: why does this work? num_out_bits is not binary? I tried locally for an int with 0 but it throws error for me.

compressors/tANS.py
```diff
@@ -0,0 +1,380 @@
...
+        # update the state
+        for s in data_block.data_list:
+            state, symbol_bitarray = self.encode_symbol(s, state)
+            encoded_bitarray = symbol_bitarray + encoded_bitarray
+
+        # Finally, pre-pend binary representation of the final state
+        encoded_bitarray = uint_to_bitarray(state, self.params.NUM_STATE_BITS) + encoded_bitarray
+
+        # add the data_block size at the beginning
+        # NOTE: rANS decoding needs a way to indicate where to stop the decoding
```
tpulkit, line 178: tANS decoding

compressors/tANS.py
```diff
@@ -0,0 +1,380 @@
...
+        # Finally, as a sanity check, ensure that the end state should be equal to the initial state
+        assert state == self.params.INITIAL_STATE
+
+        return DataBlock(decoded_data_list), num_bits_consumed
+
+
+######################################## TESTS ##########################################
+
+
+def test_check_encoded_bitarray():
```
tpulkit, line 267: Can we also explicitly test the generated tables here? Right now only checking encoded bits.
similarly might be useful to have a decoded table test. Can have a separate test_generated_tables()

tpulkit (commented):
Added minor comments! Looking great mostly. We can merge this in meeting today.

compressors/rANS.py
kedartatwawadi: @tpulkit  -> resolved by mentioning RANGE_FACTOR in the writeup

compressors/rANS.py
tpulkit: also `self.params.freqs.total_freq` can be defined in init

kedartatwawadi: Thanks ll make the change!

compressors/rANS.py
kedartatwawadi: Fixed in the writeup by @tpulkit

compressors/rANS.py
```diff
@@ -295,43 +283,19 @@ def decode_block(self, encoded_bitarray: BitArray):
...
     # test a specific example to check if the bitstream is as expected
     freq = Frequencies({"A": 3, "B": 3, "C": 2})
     data = DataBlock(["A", "C", "B"])
-    params = rANSParams(DATA_BLOCK_SIZE_BITS=5, NUM_BITS_OUT=1, RANGE_FACTOR=1)
+    params = rANSParams(freq, DATA_BLOCK_SIZE_BITS=5, NUM_BITS_OUT=1, RANGE_FACTOR=1)
 
     # NOTE: the encoded_bitstream looks like = [<data_size_bits>, <final_state_bits>,<s0_bits>, <s1_bits>, ..., <s3_bits>]
```
kedartatwawadi, line 300: @tpulkit .. ll clarify what `s0, s1,...` are

compressors/rANS.py
kedartatwawadi: Thanks ll fix. Also will add details on alternative reverse approach

compressors/tANS.py
kedartatwawadi: ll fix!

compressors/tANS.py
kedartatwawadi: `[L, H] = [M, 2^{NUM_OUT_BITS}*M - 1]` .. due to this range increases exponentially with NUM_OUT_BITS

kedartatwawadi: but ll add a comment

tpulkit: Thanks!

compressors/tANS.py
kedartatwawadi: thanks!

compressors/tANS.py
kedartatwawadi: ll look into it.. if convenient!

kedartatwawadi: Assigned to @tpulkit

compressors/tANS.py
kedartatwawadi: Done

compressors/tANS.py
kedartatwawadi: done

compressors/tANS.py
kedartatwawadi: Done

compressors/rANS.py
kedartatwawadi: done

compressors/rANS.py
kedartatwawadi: @tpulkit, I see that we only use `1 << self.params.NUM_BITS_OUT` once in the rANS code.. so keeping it that way, as no point in introducing a variable if we are not using it multiple times

kedartatwawadi: defined `self.M = self.freqs.total_freq` in the `_post_init`

compressors/tANS.py
kedartatwawadi: Actually this is done done now @tpulkit

shubhamchandak94 (approved):

tpulkit (approved):
Thanks for fixing the comments Kedar! It looks great now.

PR #61, opened 2022-10-03
Simplification to shannon codes code

kedartatwawadi:
- simplified the shannon coder code (also there might have been a small bug) 
- ran black on the `uint_coder`


tpulkit (approved):
Looks good!

PR #66, opened 2022-10-16
add zstd compressor

shubhamchandak94:


external_compressors/zlib_external.py
```diff
@@ -1,4 +1,4 @@
-"""External compressors like gzip, etc. for testing/benchmarking purposes.
+"""External compressors gzip for testing/benchmarking purposes.
```
kedartatwawadi, line 1: Please add a link to the README as well

shubhamchandak94: done

external_compressors/zstd_external.py
```diff
@@ -0,0 +1,170 @@
...
+from utils.bitarray_utils import BitArray, bitarray_to_uint, uint_to_bitarray
+from utils.test_utils import (
+    create_random_binary_file,
+    try_file_lossless_compression,
+    try_lossless_compression,
+)
+import zstandard
+
+
+def create_zstd_compressor_object(level: int):
```
kedartatwawadi, line 29: maybe add these to individual classes as static methods? (as we will never use them independently?..

kedartatwawadi (approved):
Looks good overall

external_compressors/zstd_external.py
```diff
@@ -0,0 +1,170 @@
...
+        create_random_binary_file(
+            input_file_path,
+            file_size=5000,
+            prob_dist=ProbabilityDist({44: 0.5, 45: 0.25, 46: 0.2, 255: 0.05}),
+        )
+
+        # test lossless compression
+        assert try_file_lossless_compression(
+            input_file_path, encoder, decoder, encode_block_size=1000
+        )
```
tpulkit, line 170: Should we also add a test case to check that the encoded file size matches the one generated via Zstandard if it was run stand-alone outside SCL?

tpulkit:
Looks great overall, left a minor comment on a test case for external compressors.

PR #68, opened 2022-10-16
add kth order adaptive coder

shubhamchandak94:


compressors/arithmetic_coding.py
```diff
@@ -90,6 +90,70 @@ def update_model(self, s):
                 self.freqs_current.freq_dict[s] = max(f // 2, 1)
 
 
+class AdaptiveOrderKFreqModel(FreqModelBase):
```
kedartatwawadi, line 93: does it make sense to put the `OrderKFreqModel` etc. in another file, `adaptive_orderk_arithmetic_coding.py`? so that `arithmetic_coding.py` file is cleaner, and we have a focused file dealing with the o`orderk` stuff?

tpulkit: I think maybe it makes sense to have the whole `FreqModelBase` class and the inherited classes in another file, maybe in utils? Because frequency model classes can be used with other compressors too. Or in future we can have more complicated probability models. Maybe we should have a `prob_models` file/folder?

compressors/arithmetic_coding.py
```diff
@@ -90,6 +90,70 @@ def update_model(self, s):
...
+    def __init__(self, alphabet: List, k: int, max_allowed_total_freq):
+        assert k >= 0
+        self.k = k
+        # map alphabet to index from 0 to len(alphabet) so we can use with numpy array
+        self.alphabet_to_idx = {alphabet[i]: i for i in range(len(alphabet))}
+        # keep freq/counts of (k+1) tuples, initialize with all 1s (uniform)
+        self.freqs_kplus1_tuple = np.ones([len(alphabet)] * (k + 1), dtype=int)
+        self.max_allowed_total_freq = max_allowed_total_freq
+        # keep track of past k symbols (i.e., alphabet index) seen. Initialize with all 0s.
+        self.past_k = [0] * k
```
tpulkit, line 111: Why all 0's for initialization? Does this assume something about the alphabets will always contain 0?

compressors/arithmetic_coding.py
```diff
@@ -435,3 +511,96 @@ def test_adaptive_arithmetic_coding():
         lossless_entropy_coder_test(
             encoder, decoder, freq, DATA_SIZE, encoding_optimality_precision=1e-1, seed=0
         )
+
+
+def _generate_2nd_order_markov(num_samples: int, seed: int = 0):
```
tpulkit, line 516: maybe this can be a part of the `test_utils` as there are similar functions in `test_utils` already.

compressors/arithmetic_coding.py
tpulkit, line 111: So here we are assuming that when we start calculating adaptive frequencies, we start with past context of all k symbols seen to be the 0th alphabet index symbol, correct? how do typical encoders handle this case for the initialization?

compressors/arithmetic_coding.py
```diff
@@ -435,3 +511,96 @@ def test_adaptive_arithmetic_coding():
...
+    """
+
+    DATA_SIZE = 10000
+
+    data_block = _generate_2nd_order_markov(DATA_SIZE)
+
+    # for the source as defined in _generate_2nd_order_markov, the
+    # 0th and 1st order entropy is log_2(3) [uniform] and beyond that it is 1.
+
+    print()
```
tpulkit, line 585: typo?

compressors/arithmetic_coding.py
tpulkit: Left minor comments, looks great overall. Thanks.

tpulkit:
Left minor comments, looks great overall. Thanks.

compressors/arithmetic_coding.py
shubhamchandak94: I've moved to a probability models file as suggested by Pulkit. I kept it in the compressors folder still because the models are a big part of the compressor and deserve to be there.

compressors/arithmetic_coding.py
shubhamchandak94: moved there now

compressors/arithmetic_coding.py
shubhamchandak94: yes, that's it. I added a comment. I think generally you need to do something or the other for the first few. I chose a simple method to avoid complications.

compressors/arithmetic_coding.py
shubhamchandak94: no, it was to properly format the test output. Added a comment now.

compressors/arithmetic_coding.py
```diff
@@ -244,9 +177,15 @@ class ArithmeticDecoder(DataDecoder):
     - Charles Bloom's blog: https://www.cbloom.com/algs/statisti.html#A5
     """
 
-    def __init__(self, params: AECParams, freq_base: Frequencies, freq_model_cls: FreqModelBase):
+    def __init__(
+        self, params: AECParams, freq_model_params: Frequencies, freq_model_cls: FreqModelBase
+    ):
         self.params = params
-        self.freq_model = freq_model_cls(freq_base, params.MAX_ALLOWED_TOTAL_FREQ)
+        if isinstance(freq_model_params, tuple):
```
kedartatwawadi, line 184: created an issue to simplify this.. https://github.com/kedartatwawadi/stanford_compression_library/issues/70
I will look into this later

kedartatwawadi (approved):
LG!

compressors/arithmetic_coding.py
shubhamchandak94: makes sense, I am also not happy with the current hack

PR #71, opened 2022-10-22
LZ77 implementation

shubhamchandak94:
LZ77 implementation based on a mix of gzip and zstd concepts. Works well (close to gzip performance). Includes support for window initialization (dictionary in zstd parlance).

Also fixed some bugs in the existing code, see comments in PR.

Also include Elias delta code which we use in LZ77.

TODO: include some benchmark results

TODO: go through code, add comments where needed


utils/tree_utils.py
```diff
@@ -29,7 +29,7 @@ def merge_lines(lines_1, lines_2):
             """
             lines = []
             for l1, l2 in zip(lines_1, lines_2):
-                lines.append(l1 + l2)
+                lines.append(l1 + str(l2))
```
shubhamchandak94, line 32: otherwise can't print tree for uint8 alphabet

compressors/huffman_coder.py
```diff
@@ -43,6 +44,14 @@ def build_huffman_tree(self) -> HuffmanNode:
         1. Sort the prob distribution, combine last two symbols into a single symbol
         2. Continue until a single symbol is left
         """
+        # For the special case that there is a single symbol, we simply create
```
shubhamchandak94, line 47: earlier this would just create the code with empty bitarray for this case

kedartatwawadi: great find. Although, this was not the reason for Q3 issues right?

shubhamchandak94: no, but I got some errors due to this

core/data_encoder_decoder.py
```diff
@@ -55,9 +55,6 @@ def encode(self, data_stream: DataStream, block_size: int, encode_writer: Encode
             encode_writer (EncodedBlockWriter): the writer used to write encoded bitarrays
         """
 
-        # reset the state
-        self.reset()
```
shubhamchandak94, line 59: removed reset here and in decoder because that deleted the initialized window in LZ77. I think doesn't make sense to reset right after creation anyway. Please correct me if I am missing anything.

compressors/elias_delta_uint_coder.py
```diff
@@ -0,0 +1,164 @@
...
+2. Let N = floor(log_2 (Y)) [this is 1 less than the number of bits in binary repr of Y]. Note N>=0
+3. Let M = N+1 so M > 0
+4. Let L = floor(log_2 (M)) [this is 1 less than the number of bits in binary repr of M].
+5. Write L zeros, followed by
+6. The L+1 bit binary representation of M(=N+1) [note this always begins with a 1], followed by
+7. All but the leading bit of Y(=X+1) (N bits) [leading bit is always 1 so we skip]
+
+Decoding is quite simple:
+- first we read the value of L by finding the number of 0s before a 1
+- then we read L+1 bits to compute M
```
kedartatwawadi, line 17: why did we have to use Elias delta coder? is the Golomb code inefficient or has different probabilities?

compressors/huffman_coder.py
kedartatwawadi: thanks!

core/data_encoder_decoder.py
```diff
@@ -134,9 +131,6 @@ def decode(self, encode_reader: EncodedBlockReader, output_stream: DataStream):
             output_stream (DataStream): DataStream object to write decoded blocks of data
         """
 
-        # reset the state
```
kedartatwawadi, line 137: The reason to call `reset` at the beginning of `encode` and `decode` functions was so that when we call the same 
`Encoder.encode` on another input, the state maintained is re-freshed. 

As you remove the call to reset, the onus is on the user to reset the encoder (in case the user wants to). This is probably better. Can you fix the comments in the `reset` function.. as it is still saying in the NOTE that `reset` functions gets called before encode/decode.. maybe say that the user can call reset if he/she wants to

utils/tree_utils.py
kedartatwawadi: thanks for the fix

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+This is a simplified LZ77 encoder deriving ideas from: 
+- LZ77 modern implementation: https://glinscott.github.io/lz/index.html 
+- DEFLATE: https://www.rfc-editor.org/rfc/rfc1951 
+- ZSTD: https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md
+
+The encoder has 2 parameters: 
+- min_match_length: this determines the minimum
+allowed match length. The same length is also used for indexing into the dict.
+Very low values lead to slow execution. In terms of compression, low values can
+lead to suboptimal matches but high values can lead to missed matches. So there
```
kedartatwawadi, line 17: why should low values lead to sub-optimal matches? (because there are too many?)

kedartatwawadi: (I see, because of the next paramter)

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+Defaults to 64.
+
+The algorithm: 
+
+- keep a window that stores everything the encoder has seen (not reset across blocks
+ unless reset called). 
+- keep a dict that stores a map from tuples of length of min_match_length to the 
+positions where the tuple is seen. 
+- to find a match during parsing, we look up in the dict and then find the longest
+match among the most recent max_num_matches_considered. If no match at a position, we
```
kedartatwawadi, line 31: If no match at a position, we search for matches starting at next .. should we add that, if no match, that symbol is stored as a literal?

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+search for matches starting at next. We do this until we reach the end of the block
+and don't find anything.
+- the streams generated by parsing are as follows:
+1. Literals - when no matches found for a position we put in literals. This can also include
+the very end of the block which is unmatched.
+2. LZ77 sequences - this is a triple of literal_counts, match_length and match_offset. The idea 
+is that when we start looking for a match starting at pos_in_window, but only find a match starting
+at match_start_pos, we set literal_counts to match_start_pos-pos_in_window. The match length and
+offset have their usual meaning. Note that as in gzip/zstd we allow match_length to be greater than
```
kedartatwawadi, line 40: maybe good to add what match_offset is `cur_pos - match_start_pos` (is that right?.. as it is not completely obvious)

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+or equal to the offset, denoting overlapping matches. Decoding this needs to be done a bit carefully
+since a simple memcpy doesn't work. Finally, note that the sequences do not encode the last part
+of the block that has only literals and no matches. After executing all the sequences, the decoder
+simply copies over any leftover literals to the output.
+
+Entropy encoding:
+- Literals: encoded with Huffman based on empirical distribution. The counts are stored using
+Elias Delta code.
```
kedartatwawadi, line 48: why stored using Elias Delta btw?

compressors/elias_delta_uint_coder.py
shubhamchandak94: Golomb is more suited for geometric distribution where the probabilities fall really fast. Here we often get larger values and wanted something that uses around log n bits to encode n, so more of a 1/n type probability decay. I looked the gzip and zstd codes and they both were using this sort of decay.

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+
+Given the above, note that you could still reset the encoder in order after a
+few blocks to limit memory usage.
+
+Benchmarks on a few files from https://github.com/nemequ/squash-corpus and 
+https://corpus.canterbury.ac.nz/descriptions/#cantrbry (plus a few handmade).
+
+All sizes are in bytes.
+
+| File                                | raw size | lz77 size (this) | gzip size |
```
kedartatwawadi, line 75: maybe call this scl-lz77 :)

compressors/lz77.py
shubhamchandak94: just a design decision, I just needed something to encode a bunch of non-negative integers efficiently. It uses 1 bit to represent 0s so that's also nice

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+from core.encoded_stream import EncodedBlockReader, EncodedBlockWriter
+from core.prob_dist import ProbabilityDist
+from utils.bitarray_utils import BitArray, bitarray_to_uint, uint_to_bitarray
+from utils.test_utils import (
+    create_random_binary_file,
+    try_file_lossless_compression,
+    try_lossless_compression,
+)
+
+ENCODED_BLOCK_SIZE_HEADER_BITS = (
```
kedartatwawadi, line 107: @shubhamchandak94  -> ok for now, but maybe add this in a 

```
@dataclass
class LZ77Params:
          ENCODED_BLOCK_SIZE ... 
          

```

And use this by default in the encoder/decoder.

kedartatwawadi: Can you please add an issue to fix this?

kedartatwawadi: Actually nvm, you have these parameters as separate inputs.

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+        Args:
+            encoded_bitarray (BitArray): encoded bit array
+        """
+        # first encode the literals with empirical Huffman code
+        counts = DataBlock(literals).get_counts()
+        if len(counts) > 0:
+            prob_dict = ProbabilityDist.normalize_prob_dict(counts).prob_dict
+            # let's sort prob_dist by the alphabet to make sure we get exact same ordering
+            # during decompression as well, which makes sure it's the same Huffman tree!
+            prob_dist_sorted = ProbabilityDist({i: prob_dict[i] for i in sorted(prob_dict)})
```
kedartatwawadi, line 238: should we add this in the Huffman code construction, so that the Huffman coder has less issues?

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+            prob_dist=ProbabilityDist({44: 0.5, 45: 0.25, 46: 0.2, 255: 0.05}),
+        )
+
+        # test lossless compression
+        assert try_file_lossless_compression(
+            input_file_path, encoder, decoder, encode_block_size=1000
+        )
+
+
+# Provide a simple CLI interface below for convenient experimentation
```
kedartatwawadi, line 680: define  the parser args after ` if __name__ == "__main__":`
As if we import multiple such files, it will cause issues.

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+            block_size (int): choose the block size to be used to call the encode function
+        """
+        # call the encode function and write to the binary file
+        with Uint8FileDataStream(input_file_path, "rb") as fds:
+            with EncodedBlockWriter(encoded_file_path) as writer:
+                self.encode(fds, block_size=block_size, encode_writer=writer)
+
+
+class LZ77Decoder(DataDecoder):
+    def __init__(self, window_initialization: List = None):
```
kedartatwawadi, line 388: should this be called `initial_window` instead of `window_initialization`? (minor)

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+        Args:
+            encoded_bitarray (BitArray): encoded bit array
+        """
+        # first get the encoded block size so we can pick out the block for decoding
+        encoded_block_size_bitarray = encoded_bitarray[:ENCODED_BLOCK_SIZE_HEADER_BITS]
+        encoded_block_size = bitarray_to_uint(encoded_block_size_bitarray)
+        encoded_bitarray = encoded_bitarray[
+            ENCODED_BLOCK_SIZE_HEADER_BITS : ENCODED_BLOCK_SIZE_HEADER_BITS + encoded_block_size
+        ]
+        # decoded the combined list with literal counts, match lengths and offsets
```
kedartatwawadi, line 412: decoded -> decode

compressors/lz77.py
shubhamchandak94: I think it's better to make it the caller's responsibility and perhaps adding a note in the Huffman coder. In retrospect this seems a very obvious thing, I think the confusion is caused by Python's dict maybe (if we were passing a list I feel this would be clearer).

A couple reasons to not add this to Huffman:
- might cause trouble if the type is not sortable
- this issue is not specific to Huffman, it applies also to arithmetic coder, range coder etc.

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+
+    # matches here are (first one is overlapping, last one picks longer match which is not the most recent match for 3-tuple):
+    # 0, 0, [1, 1, 1, [1,] 1, 1, 1] 0, 0, 1, 1, 1, 255, 254, 255, 254, 255, 254, 255, 2, 0, 0, 1, 1, 1, 1, 44
+    # [0, 0, 1, 1, 1,] 1, 1, 1, [0, 0, 1, 1, 1,] 255, 254, 255, 254, 255, 254, 255, 2, 0, 0, 1, 1, 1, 1, 44
+    # 0, 0, 1, 1, 1, 1, 1, 1, 0, 0, 1, 1, 1, [255, 254, 255,] 254, [255, 254, 255,] 2, 0, 0, 1, 1, 1, 1, 44
+    # [0, 0, 1, 1, 1, 1,] 1, 1, 0, 0, 1, 1, 1, 255, 254, 255, 254, 255, 254, 255, 2, [0, 0, 1, 1, 1, 1], 44
+
+    expected_lits = [255, 254, 255, 254, 2, 44]
+    expected_seqs = [
+        LZ77Sequence(0, 4, 3),
```
kedartatwawadi, line 614: are overlapping matches ok?

shubhamchandak94: yes

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+        while start_pos_1 + match_length < len(self.window) and start_pos_2 + match_length < len(
+            self.window
+        ):
+            if self.window[start_pos_1 + match_length] != self.window[start_pos_2 + match_length]:
+                break
+            else:
+                match_length += 1
+        return match_length
+
+    def lz77_parse_generate_sequences(self, data_block: DataBlock):
```
kedartatwawadi, line 285: minor -> should this be named `lz77_parse_and_generate_sequences`? (as currently it seems like you are parsing the generated sequences)

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+
+    def decode_literals(self, encoded_bitarray: BitArray):
+        """Perform entropy decoding of the literals and return the literals
+        and the number of bits consumed.
+
+        Args:
+            encoded_bitarray (BitArray): encoded bit array
+        """
+        num_bits_consumed = 0
+        # first read the size of the counts encoding
```
kedartatwawadi, line 445: maybe call this symbol counts (or frequencies), so that it is clear what counts these are..

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+        num_bits_consumed += ENCODED_BLOCK_SIZE_HEADER_BITS
+        # decode literals with Huffman
+        decoded_literals, num_bits_consumed_literals = HuffmanDecoder(prob_dist).decode_block(
+            encoded_bitarray[num_bits_consumed : num_bits_consumed + literals_huffman_encoding_size]
+        )
+        assert literals_huffman_encoding_size == num_bits_consumed_literals
+        num_bits_consumed += literals_huffman_encoding_size
+        return decoded_literals.data_list, num_bits_consumed
+
+    def execute_lz77_sequences(self, literals: List, lz77_sequences: List[LZ77Sequence]):
```
kedartatwawadi, line 474: why is this function called `execute`.. ? 
Is this legacy, and to be consistent with other implementations?

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+    def encode_lz77_sequences(self, lz77_sequences: LZ77Sequence):
+        """Perform entropy encoding of the LZ77 sequences and return the encoded bitarray.
+
+        We simply use the Elias Delta code for encoding all three, but first we preprocess a bit to
+        optimize the encoding.
+
+        Args:
+            encoded_bitarray (BitArray): encoded bit array
+        """
+        literal_counts = [l.literal_count for l in lz77_sequences]
```
kedartatwawadi, line 204: surprised that just Elias-Gamma coding for all the params of the `lz77_sequences` turns out to be not-so-bad for encoding.

compressors/lz77.py
```diff
@@ -0,0 +1,705 @@
...
+                in order of recency and the longest match among the max_num_matches_considered
+                is chosen. Defaults to 64.
+            window_initialization (List, optional): initialize window (this is like side information
+                or dictionary in zstd parlance). The same initialization should be used for the decoder.
+
+        """
+        self.min_match_length = min_match_length
+        self.max_num_matches_considered = max_num_matches_considered
+        self.window = []
+        self.dict = (
```
kedartatwawadi, line 148: maybe name this variable a bit better. `self.dict` is not very descriptive

kedartatwawadi: substring_dict?

kedartatwawadi (approved):
great job getting all of this working! 
Added some suggestions for function names/var names and more clarification. 

I think the code is written quite well, and relatively easy to understand.

compressors/lz77.py
kedartatwawadi: Cool, is Elias-gamma used by gzip as well?

compressors/lz77.py
kedartatwawadi: sounds good, yeah we should make the prob_dist an orderedDict more explicitly (`dict` is ordered nowadays, but is not very explicit)

compressors/lz77.py
shubhamchandak94: yes

compressors/lz77.py
shubhamchandak94: zstd parlance

compressors/lz77.py
shubhamchandak94: I was inspired by the deflate standard which uses a fixed encoding.

compressors/lz77.py
shubhamchandak94: It's actually `match_start_pos-best_candidate_start_pos`, will to the description.

compressors/lz77.py
shubhamchandak94: they use a fixed code as shown in the deflate rfc whose lengths looked similar to elias delta (basically around log n for encoding n). So instead of using a fixed table, I just used this (there is a slight inefficiency but probably not too bad)

compressors/lz77.py
shubhamchandak94: added clarification for this

compressors/lz77.py
shubhamchandak94: done now

compressors/lz77.py
shubhamchandak94: done now

compressors/lz77.py
shubhamchandak94: done now

compressors/lz77.py
shubhamchandak94: done now

compressors/lz77.py
shubhamchandak94: done now

compressors/lz77.py
shubhamchandak94: done now

PR #75, opened 2022-10-29
cleanup the fixed bitwidth compressor

kedartatwawadi:
1. cleanup
2. create an encoder which uses pickle for alphabet encoding/decoding
3. fix the func which returns a list from numpy array, as if it is not a list, the pickled side is ~10x larger


external_compressors/pickle_external.py
```diff
@@ -0,0 +1,72 @@
...
+        length_encoding = uint_to_bitarray(len_pickled, bit_width=self.length_bitwidth)
+        return length_encoding + pickled_bits
+
+
+class PickleDecoder(DataDecoder):
+    def __init__(self, length_bitwidth=32):
+        self.length_bitwidth = length_bitwidth
+
+    def decode_block(self, bitarray: BitArray):
+        length_bitwidth = 32
```
shubhamchandak94, line 40: why set to 32 here and not use self.length_bitwidth?

compressors/fixed_bitwidth_compressor.py
```diff
@@ -161,7 +185,23 @@ def test_fixed_bitwidth_encode_decode():
     assert codelen == len(data_list) * 2 + alphabet_bits
 
 
-def test_fixed_bitwidth_file_encode_decode():
+def test_fixed_bitwidth_encode_decode():
```
shubhamchandak94, line 188: test also non-character alphabet?

external_compressors/pickle_external.py
kedartatwawadi: thanks! ll fix

compressors/fixed_bitwidth_compressor.py
kedartatwawadi: as in? can you elaborate?

shubhamchandak94: Just meant test on integer or something.

shubhamchandak94 (approved):

PR #76, opened 2022-10-30
Add vector quantization code

kedartatwawadi:
NOTE-> depends on the other 2 prs, so lets merge them first

### DIFF
--- a/lossy_compressors/vector_quantizer.py
+++ b/lossy_compressors/vector_quantizer.py
@@ -1,0 +1,223 @@
+from typing import List
+from compressors.fixed_bitwidth_compressor import (
+    FixedBitwidthDecoder,
+    FixedBitwidthEncoder,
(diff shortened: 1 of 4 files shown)


lossy_compressors/vector_quantizer.py
```diff
@@ -0,0 +1,223 @@
...
+    avg_distortion_arr = []
+    for it in range(max_iter):
+        # define intermediate tensors and vals
+        counts = np.ones(num_vectors)
+        centroids = codebook.copy()  # current centroid
+        sum_distances = 0
+
+        # go over the input data and average
+        for data_vector in data_npy:
+            min_ind, dist = find_nearest(codebook, data_vector, distortion_func)
```
kedartatwawadi, line 69: @tpulkit  -> min is found wrt to the codebook, which is updated using the centroids only at the end

PR #83, opened 2022-12-05
Kedar/scl package

kedartatwawadi:
making the repo into a package. Maybe lets merge this after the course? 
Some more things:
- might be good to add some "plug-n-play" of Huffman, Adaptive Arithmetic and LZ which don;t requite any input
- Might be good to see how to make the repo a bit less bloated (maybe remove `ProbDist` object etc. and use simple ordered dict to represent it? (Need to think about it). Usually better to have less custom classes. Some classes like encoded stream, data stream might still be useful though :-|


.github/workflows/black.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
+        with:
+            options: "--check --line-length 100 --verbose"
+            src: "./stanford_compress"
+            version: "~= 22.3"
```
kedartatwawadi, line 20: black version needs to be at least 22.3, due to some bug

.github/workflows/black.yml
```diff
@@ -0,0 +1,20 @@
+
+
+name: Lint
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
```
kedartatwawadi, line 8: Checks if black was run or not, and errors if not

.github/workflows/python-app.yml
```diff
@@ -23,15 +23,8 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 pytest
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
kedartatwawadi, line 27: we don't use flake, so just removed that.. added a separate action for black formatting

README.md
```diff
@@ -8,50 +8,55 @@ This library is currently being used for the course [EE274: Data Compression cou
...
     ```
     conda create --name myenv python=3.8.2
     conda activate myenv
-    python -m pip install -r requirements.txt
+    
     ```
-- Add path to the repo to `PYTHONPATH`:
+- Install the `scl` package
     ```
-    export PYTHONPATH=$PYTHONPATH:<path_to_repo>
```
kedartatwawadi, line 42: no need to export pythonpath etc.

scl/core/prob_dist.py
```diff
@@ -60,7 +59,6 @@ def cumulative_prob_dict(self):
         return cum_prob_dict
 
     @property
-    @cache
```
kedartatwawadi, line 63: removed @cache decorator as it has unpredicable behavior. 
TODO -> make cumul prob_dist just members instead of `@property` as that is recomputing things

README.md
```diff
@@ -2,56 +2,62 @@
...
 
-NOTE -> the tests in each file should be helpful as a "usage" example of each of the compressors. More details are also available in the lecture [slides](https://stanforddatacompressionclass.github.io/Fall22/lectures/), and the project wiki
 
+NOTE -> the tests in each file should be helpful as a "usage" example of each of the compressors. More details are on the project wiki.
 
 
 ## Getting started
 - Create conda environment and install required packages:
     ```
     conda create --name myenv python=3.8.2
```
shubhamchandak94, line 42:
```(base) ➜  code conda create --name scltestenv python=3.8.2
Collecting package metadata (current_repodata.json): done
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:

  - python=3.8.2

```

whereas python=3.8 works for me

shubhamchandak94: maybe because there's a newer 3.8 available?

README.md
```diff
@@ -2,56 +2,62 @@
...
     conda activate myenv
-    python -m pip install -r requirements.txt
+    
     ```
-- Add path to the repo to `PYTHONPATH`:
+- Install the `scl` package
     ```
-    export PYTHONPATH=$PYTHONPATH:<path_to_repo>
+    python install -e . #install the package in a editable mode
```
shubhamchandak94, line 48:
```
(scltestenv) ➜  scl_test git:(kedar/scl_package) python install -e .
/usr/local/bin/python3: can't open file '/Users/shubhamchandak/code/scl_test/install': [Errno 2] No such file or directory
```

shubhamchandak94: not sure what to do

shubhamchandak94: for me `python setup.py install` worked but not sure that's the right thing to do

README.md
```diff
@@ -2,56 +2,62 @@
...
     ``` 
 
 - **Run unit tests**
 
   To run all tests:
     ```
-    find . -name "*.py" -exec py.test -s -v {} +
+    find  scl -name "*.py" -exec py.test -s -v {} +
```
shubhamchandak94, line 55: for the tests, do we still need to set the pythonpath (the test below suggests so)

README.md
shubhamchandak94: after python setup.py install, when I opened python interpreter and imported scl it worked well (I tested creating a datablock and computing entropy). But then I went to a different directory and tried import scl but that errored. Is that expected?

README.md
kedartatwawadi: Not sure, python version is not critical for us btw:
maybe `conda update --all` might work. @tpulkit might know more about conda.

README.md
kedartatwawadi: did you do it from the `stanford_compression_library` dir? seems to work for me.

README.md
kedartatwawadi: Actually we dont need to.. I removed the un-necessary pythonpath

README.md
kedartatwawadi: Once the package is added to the python env, we should be able to do that from other directories as well.. I am guessing that scl package hasnt been installed properly as of now.

README.md
tpulkit: Fixed. 3.8 works, 3.8.2 is not available.

tpulkit:
Looks good and works!

shubhamchandak94 (approved):

tpulkit (approved):

PR #86, opened 2023-04-03
readme: fix link and typo

tao-j:


README.md
```diff
@@ -52,12 +51,12 @@ NOTE -> the tests in each file should be helpful as a "usage" example of each of
...
   To run all tests:
     ```
-    find  scl -name "*.py" -exec py.test -s -v {} +
+    find scl -name "*.py" -exec py.test -s -v {} +
     ```
 
   To run a single test
   ```
-  py.test -s -v scl/core/data_stream_tests.py
+  py.test -s -v scl/core/data_stream.py
```
kedartatwawadi, line 59: Thanks for the changes! can you please revert line 59, as that seems to be correct

tao-j: All tests passed but single test complained about not finding the scl module. Possible  venv issue. 

I just reverted the change. In case I need to use/write single test again I will revist and verify.

kedartatwawadi (approved):
looks good!

PR #96, opened 2023-10-27
Update LZ77 entropy coder + Add LZ77 sliding window compression

shubhamchandak94:
# Update LZ77 entropy coder

I updated the LZ77 entropy to use Huffman coding on the lengths etc. as well. Now it's quite similar to what gzip and zstd do. I also refactored the entropy coder which will be required for the sliding window version of LZ77 that is in progress. I also tried to do 0th order AAC instead of Huffman but that was way too slow and also didn't give appreciable size difference. The performance on very small files has a small hit due to slight extra overhead but that shouldn't affect any conceptual understanding from experiments.

This adds:
- empirical Huffman coder for alphabet [0,n-1]
- LogScaleBinnedIntegerEncoder (bin an integer range logarithmically, Huffman code the bin index and plain encode the position within bin)

The performance change (note impact is pretty big on some files, smaller on others but overall we always improve or stay very close). Note also how we are now much closer to gzip!

```
| File                                | raw size | scl-lz77 size    | gzip size |scl-lz77 size (old)|
|-------------------------------------|----------|------------------|-----------|-------------------|
| bootstrap-3.3.6.min.css             |121260    |20126             |19747      |23526              |
| eff.html                            |41684     |10755             |9670       |10935              |
| zlib.wasm                           |86408     |41156             |37448      |43471              |
| jquery-2.1.4.min.js                 |84345     |31322             |29569      |34447              |
| random.bin (random bytes)           |1000000   |1005050           |1000328    |1004970            |
| repeat_As.txt                       |1000000   |578               |1004       |208                |
| kennedy.xls                         |1029744   |210182            |204004     |330252             |
| alice29.txt                         |152089    |54106             |54416      |63376              |
```

# Add LZ77 sliding window compression 

This adds a sliding window version of LZ77 that is much closer to real-life LZ77 implementations. Another major addition is to separate out the match finder into a class allowing easy experimentation with different match finders. See the doc comments at the start of lz77_sliding_window.py for more details.


scl/compressors/lz77.py
```diff
@@ -128,6 +124,310 @@ class LZ77Sequence:
...
+            encoded_bitarray[num_bits_consumed : num_bits_consumed + huffman_encoding_size]
+        )
+        assert huffman_encoding_size == num_bits_consumed_vals
+        num_bits_consumed += huffman_encoding_size
+        return decoded_vals, num_bits_consumed
+
+
+class LogScaleBinnedIntegerEncoder(DataEncoder):
+    """
+    Encodes a list of non-negative integers by binning in log scale and then
```
chachachaudhary274, line 214: The magnitude of difference (residual) can be between 0 - 2^logarithm, which can be a large range. Maybe comment on what kind of distributions is this encoding good. (exponential?)

chachachaudhary274 (approved):

tpulkit (approved):
Great PR!

scl/compressors/lz77.py
shubhamchandak94: we actually use a fixed length code for that

PR #99, opened 2025-01-08
delete redundant code

DayuanJiang:
Removes redundant heapify calls within the `build_huffman_tree` function. The `heapq.heappush(node_heap, combined_node)` operation already maintains the heap property, making the initial heapify call in each loop unnecessary.


chachachaudhary274 (approved):
Thanks a lot for the contribution

